PR #68, opened 2025-12-14
Rerwite Flatten Move program to Squin

### LINKED ISSUE #9: Rewrite from move language to Squin.
Depends on the completion of #6. This can be broken into two parts. 

- [ ] Noise-less rewrite rule from Move to Squin. 
- [ ] Noisy rewrite rule from Move To Squin

weinbe58:
this PR implements the rewrites and pass that can rewrite a compiled move program back to Squin including:
1. tracking atoms and inserting the appropriate gates
2. A rewrite that inserts move noise based on the input lanes. 
closes #9


weinbe58:
Found a bug in the rewrites but I need to commit this.

PR #88, opened 2025-12-17
Refactor Measurement semantics in `move` dialects. 

weinbe58:
This PR implements a new measurement future semantics that more mimics the current proposed hardware implementation.

### DIFF
--- a/src/bloqade/lanes/analysis/placement/lattice.py
+++ b/src/bloqade/lanes/analysis/placement/lattice.py
@@ -77,7 +77,17 @@ class ConcreteState(AtomState):
 @final
 @dataclass
 class ExecuteCZ(ConcreteState):
+    """Defines the state representing the placement of
+    atoms before/after executing CZ gate pulse.
+
+    NOTE: you can specify multiple entnangling zones to be active
+    in a single ExecuteCZ state in cases where there are multiple entangling
+    zones that can be used in parallel.
+
+    """
+
     active_cz_zones: frozenset[ZoneAddress]
+    """The set of CZ zones that need to execute for this round of CZ gates."""
 
     @classmethod
     def from_concrete_state(
@@ -101,7 +111,13 @@ class ExecuteCZ(ConcreteState):
 @final
 @dataclass
 class ExecuteMeasure(ConcreteState):
-    """A state representing measurement placements."""
+    """A state representing measurement placements.
+
+    NOTE: Depending on the placement of the atoms you may need to specify
+    which atoms are measured by which zone. This is done via the zone_maps field, such that
+    `zone_maps[i]` gives the zone that measures the ith qubit.
+
+    """
 
     zone_maps: tuple[ZoneAddress, ...]
 


src/bloqade/lanes/layout/arch.py
```diff
@@ -36,6 +36,10 @@ class ArchSpec(Generic[SiteType]):
     """tuple of all words in the architecture. words[i] gives the word at word address i."""
     zones: tuple[tuple[int, ...], ...]
     """A tuple of zones where a zone is a tuple of word addresses and zone[i] gives the ith zone."""
+    measurement_zones: frozenset[int]
+    """Set of zone ids that support measurements."""
+    cz_gate_zones: frozenset[int]
```
nkanazawa1989:
```suggestion
    entangling_zones: frozenset[int]
```
`ArcSpec` should be hardware agnostic and future architecture may implement non-cz gate.

src/bloqade/lanes/layout/arch.py
```diff
@@ -36,6 +36,10 @@ class ArchSpec(Generic[SiteType]):
     """tuple of all words in the architecture. words[i] gives the word at word address i."""
     zones: tuple[tuple[int, ...], ...]
     """A tuple of zones where a zone is a tuple of word addresses and zone[i] gives the ith zone."""
+    measurement_zones: frozenset[int]
```
nkanazawa1989: What's unclear to me is translation between "zone" and "mode". Seems this conversion is pretty arbitrary because mode in flair only sees the SLM site coordinates, and we can get the information from `Word.all_positions()` -- they can be mapped via physical coordinates. 

Currently this mapping is implicit and on responsibility of a developer who writes rewrite rules. This also makes the rewrite rule not portable. I would define data structure to manage this mapping in this spec to make it more explicit.

test/rewrite/test_transversal.py
```diff
@@ -83,59 +83,6 @@ def test_simple_rewrite(
     assert result.has_done_something is has_done_something
 
 
-def test_get_measurement_result():
```
nkanazawa1989: Why these tests are completely gone? If we plan to readd in future we can also use `@pytest.mark.skip`.

src/bloqade/lanes/rewrite/place2move.py
```diff
@@ -105,12 +111,13 @@ def rewrite_Statement(self, node: ir.Statement) -> RewriteResult:
         if not isinstance(state_after, placement.ConcreteState):
             return RewriteResult()
 
-        zone_addresses = self.move_heuristic.compute_zone_addresses(
-            tuple(state_after.layout[i] for i in node.controls + node.targets)
-        )
+        location_addresses = tuple(state_after.layout[i] for i in node.qubits)
+        if not self.move_heuristic.all_in_cz_gate_zone(location_addresses):
+            return RewriteResult()
```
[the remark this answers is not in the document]
weinbe58: yeah true.

src/bloqade/lanes/rewrite/place2move.py
```diff
@@ -15,17 +16,22 @@
 @dataclass
 class MoveSchedulerABC(abc.ABC):
     arch_spec: ArchSpec
+    measurement_zone_address: ZoneAddress = field(init=False, default=ZoneAddress(0))
```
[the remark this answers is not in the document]
weinbe58: Actually I udpated the placement analysis to provide which zones need to be used in the rewrite which now makes the rewrite a lot cleaner.

PR #99, opened 2026-01-11
Add state to specialization

### LINKED ISSUE #97: Refactor Specialization to be compatible with new semantics in `move`.
After #90 The new statements in the `move` dialect can consume and/or emit a global state value. As such the old specialization statements are not compatible. Here we should make sure that the statements can consume and emit the global state value. Currently this is breaking the pipeline. 

We need to:

1. Update the dialect [here](https://github.com/QuEraComputing/bloqade-lanes/blob/main/src/bloqade/lanes/arch/gemini/logical/stmts.py) to consume and emit a state value. You can simply import the base statement class `StatefulStatement` found in the `move` dialect. 
2. Update [rewrites](https://github.com/QuEraComputing/bloqade-lanes/blob/main/src/bloqade/lanes/arch/gemini/logical/rewrite.py) from the `move` dialect to this specialization dialect to account for the new state value
3. Update the downstream compiler passes to account for the new dialect.

nkanazawa1989:
This PR adds handling of state attribute to Gemini specializations. I assume the atom state data structure is preserved in Gemini.

I realized some upstream package release updated uv.lock file. Because our team often configures uv to use an internal artifactory as index, pre-commit should not implicitly update the lock file. It’s better for us to run `uv lock --no-config` manually, otherwise we may accidentally use the artifactory as source.

In ad1cf0b740f627458fb76b245761dfc3cb059751 I modified the pre-commit file. Strictly speaking this is not a part of this PR. I can create separate one if you prefer @weinbe58 

Fix #97

### DIFF
--- a/src/bloqade/lanes/arch/gemini/logical/rewrite.py
+++ b/src/bloqade/lanes/arch/gemini/logical/rewrite.py
@@ -43,7 +43,8 @@ class RewriteMoves(rewrite_abc.RewriteRule):
         if move_type is MoveType.SITE:
             node.replace_by(
                 stmts.SiteBusMove(
-                    y_mask_ref,
+                    current_state=node.current_state,
+                    y_mask=y_mask_ref,
                     word=word,
                     bus_id=bus_id,
                     direction=direction,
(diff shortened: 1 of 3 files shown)


weinbe58:
I'll take a look at the demo after merging!

PR #105, opened 2026-01-16
Add simple keybindings to the debugger interface

lsschmid:
This PR registers simple keybindings (arrow keys) to the buttons to make it easier to step through the movements.


weinbe58 (changes requested):
Good addition overall but I was thinking of using `esc` instead of the down arror if that is OK with you?

PR #156, opened 2026-02-12
Added _lane_map functionality to ArchSpec and testing in test_arch.py…

Jason-Ludmir:
Also added an always true merge heuristic function to upstream.py for testing the logical, and included the fix to the upstream.py rewrites for NativeToPlace w/the no_raise flag set to false.

Also included testing for the new ArchSpec _lane_map and get_lane_address function.


weinbe58 (approved):
LGTM, just that one small line change and it should be good to merge!

PR #162, opened 2026-02-17
NoHome implementation to nearest left column

Jason-Ludmir:


demo/pipeline_details.py
weinbe58: Can you remove the updates to this file? If you want to have a script you use to debug just create a `debug` folder locally and store it there, I have configured the `.gitignore` to ignore any `debug` folder in the repo.

PR #397, opened 2026-04-01
fix(search): block inert occupied destinations during traversal

### LINKED ISSUE #396: fix(search): account for inert occupied locations in physical move legality
## Problem

Physical search can produce move plans that collide at runtime because inert (non-participating) atoms are excluded from search occupancy checks.

The search code correctly excludes occupied destinations, but its `occupied` set is derived from the active search configuration only. In static placement contexts, that configuration omits inert atoms, so moves into inert-occupied sites are considered legal by search and later fail during atom interpretation with `failure_reason: "collision"`.

## Symptoms

- AVO/evaluation outputs include invalid candidates with:
  - `"valid": false`
  - `"failure_reason": "collision"`
- Entropy tree reports successful expansions/goals, yet emitted move IR collides in atom simulation.

## Root Cause

Occupancy scope mismatch:

- Search legality/scoring/generation use `node.occupied_locations`
- `node.configuration` contains active qubits only for the static block
- Inert atoms live in `ConcreteState.occupied` and were not represented in search occupancy

So search “never move to occupied destination” was true only for active occupancy, not full physical occupancy.

## Reproduction (example)

For generated kernel batch (`generated_kernels_seed0128.json`), known failing cases included indices `2`, `6`, and `7` before fix:
- search reported created children / goals
- atom interpreter detected collisions in emitted move layers

## Proposed / Implemented Fix

Introduce immutable blocked occupancy into search:

1. Add `blocked_locations` to `ConfigurationTree`
2. Build tree with blockers from `ConcreteState.occupied`
3. Use effective occupied set everywhere legality/scoring/fallback path checks occur:
   - `node.occupied_locations | tree.blocked_locations`

## Files involved

- `python/bloqade/lanes/search/tree.py`
- `python/bloqade/lanes/heuristics/physical_movement.py`
- `python/bloqade/lanes/search/scoring.py`
- `python/bloqade/lanes/search/generators.py`
- `python/bloqade/lanes/search/traversal/entropy_guided.py`

## Verification

Targeted regression tests cover:
- collision detection against blocked destinations
- valid lane filtering excluding blocked destinations
- scoring excludes blocked-destination rides
- placement strategy threads idle occupancy into tree blockers

Additionally, previously colliding kernels were rechecked and no longer collide in direct compile + atom-interpreter validation.

Jason-Ludmir:
Treat idle qubit locations as immutable blockers in search legality and scoring so generated move plans do not collide with inert atoms at runtime.

closes #396

### DIFF
--- a/demo/simulator_device_demo.py
+++ b/demo/simulator_device_demo.py
@@ -54,9 +54,9 @@ task = GeminiLogicalSimulator().task(main)
 
 # run simulation with and without noise
 print("Running simulation with noise...")
-future = task.run_async(100000)
+future = task.run_async(10000)
 print("Running simulation without noise...")
-future_wo_noise = task.run_async(100000, with_noise=False)
+future_wo_noise = task.run_async(10000, with_noise=False)
 
 result_wo_noise = future_wo_noise.result()
 result = future.result()
--- a/python/bloqade/lanes/arch/gemini/impls.py
+++ b/python/bloqade/lanes/arch/gemini/impls.py
@@ -112,11 +112,13 @@ def _calc_hypercube_word_path_dict(words: tuple[Word, ...]):
 
     path_dict: dict[LaneAddress, tuple[tuple[float, float], ...]] = {}
 
+    buses = _hypercube_busses(hypercube_dims)
     for bus_id in range(hypercube_dims):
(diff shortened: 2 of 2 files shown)


weinbe58 (approved):
LGTM

PR #433, opened 2026-04-06
fix(search): stabilize entropy traversal continuation and ancestor-cycle handling

### LINKED ISSUE #432: search: clarify entropy traversal reliability and continuation gaps
## Problem

The entropy-guided search path has accumulated multiple reliability gaps that show up together in real workloads, especially once we start collecting multiple goal candidates.

## Symptoms Observed

- Branches can degenerate into local oscillations by revisiting branch-ancestor configurations instead of making progress.
- Traversal continuation after finding a goal is brittle: high-value frontier states may be dropped.
- Goal collection behavior and continuation policy are under-specified when multiple goal nodes are requested.
- Traversal stopping/fallback behavior can interact poorly with the above, making outcomes sensitive to exploration order.

## Why this matters

These issues reduce search quality and predictability, and make debugging difficult because behavior depends on subtle interactions among tree expansion, entropy/reversion logic, and resume-point selection.

## Scope

Track the full traversal-quality problem across search-tree expansion and entropy-guided continuation behavior, including ancestor revisit handling, multi-goal continuation semantics, and resume-point reliability.

Jason-Ludmir:
## Summary
- broaden entropy-guided traversal reliability updates across `tree`, `entropy_guided`, `search_params`, and physical movement wiring
- prevent branch-local ancestor state revisits from being re-expanded while preserving cross-branch exploration
- improve multi-goal continuation via buffered high-scoring resume states and add regression tests for `state-seen`/transposition behavior

## Related
- Refines problem scope tracked in #432
- Stacked on top of #431 (base branch: `feat/search-entropy-aod-grouping-418`)

## Test plan
- [x] `uv run pytest python/tests/search/test_tree.py -k transposition`
- [x] `uv run pytest python/tests/search/test_entropy_guided.py -k "state_seen or ancestor_revisit"`
- [x] pre-commit hooks on commit (`isort`, `black`, `ruff`, `pyright`) passed

Closes #432 


weinbe58 (changes requested):
## Code Review

Overall the search reliability improvements are heading in the right direction — the resume buffer, multi-goal collection, and candidate cache refresh are well-structured. A few issues need attention before merge.

---

## Critical

### 1. `collision.add(dst)` breaks collision detection (aod_grouping.py)

The change stores **destination** addresses in `collision_srcs`, but `is_valid_rect` checks **source** locations looked up from `pos_to_loc`. Since bus src/dst are disjoint, the collision check is now a guaranteed no-op — no rectangle will ever be rejected for collision.

This is the opposite of the intended fix. The correct version:
```python
if dst in occupied:
    collision.add(loc)  # source, not dst
```

See #439 for the full bug description.

### 2. Transposition table replaced with id-based tracking (tree.py)

`seen` is now keyed by `id(node)` instead of `config_key`. Since every `ConfigurationNode` gets a unique `id()`, this is effectively a write-only set that never deduplicates. Only direct ancestor-chain cycle detection remains via `_ancestor_with_config_key`.

Two independent branches reaching the same configuration will both be expanded. For small test cases this works, but for production problems with 10+ qubits, the search space could explode exponentially. If this is intentional (to allow path diversity), it should be documented with a note about the scaling tradeoff.

---

## Major

### 3. `_is_valid_rectangle_candidate` rejects resolved qubits (heuristic.py)

```python
if node.configuration[qid] == self.scorer.target[qid]:
    return False
```

This rejects any rectangle containing a qubit already at its target. But AOD rectangles are physical groupings — a resolved qubit's lane might be necessary for other qubits in the grid to move together. This could silently eliminate valid movesets.

### 4. Fallback stats silently overwritten (entropy_guided.py)

After `_sequential_fallback`, `nodes_expanded` and `max_depth_reached` are reset to pre-fallback values. The returned `SearchResult` doesn't account for fallback work. If intentional, this should be documented.

---

## Behavioral changes to document

These aren't bugs but change default behavior for all callers:

| Parameter | Old | New | Impact |
|---|---|---|---|
| `max_goal_candidates` | 1 | 2 | Search always finds 2 goals now |
| `w_m` (mobility weight) | 0.1 | 0.3 | 3x mobility preference in scoring |
| `max_expansions` | 300 | 1000 | 3x search budget |
| Distance heuristic | duration-based | hop-count (`lambda: 1.0`) | Changes which candidates are preferred |

### 5. `LaneAddress` construction drops `zone_id` (aod_grouping.py)

Line 115 of diff: `LaneAddress(move_type, loc.word_id, loc.site_id, bus_id, direction)` — omits `loc.zone_id`, defaulting to 0. This is fine for single-zone architectures but will silently break for multi-zone.

---

## Minor

- `_buffer_insert` does O(n) scan per insert — fine for small capacity, consider heapq if `max_goal_candidates` grows
- Stale comment in `greedy_init`: "lande" → "lane"
- `_ScoredResumeState.order` tiebreaking semantics should be documented

Jason-Ludmir:
Addressed per latest decisions:
- Fixed collision filtering in BusContext: collision set now stores source locations when destination is occupied.
- Added regression coverage for this in python/tests/search/test_greedy_generator.py (test_from_tree_collision_sources_store_sources_not_destinations).
- Kept branch-diversity semantics for transposition handling intentionally (option B): ancestor-chain revisits still return TRANSPOSITION_SEEN, but equivalent configs reached via different branches remain independently explorable. Added explicit documentation in ConfigurationTree doc/comments.
- Skipped issue (3) and (4) per decision.
- Zone-id LaneAddress constructor path was already fixed manually in this branch.

Targeted checks run locally:
- uv run pytest python/tests/search/test_greedy_generator.py -k "collision_sources_store_sources_not_destinations or is_valid_rect_collision"
- uv run pytest python/tests/search/test_tree.py -k "allows_same_config_from_different_branch or transposition_for_ancestor_revisit"

weinbe58 (approved):
LGTM

PR #462, opened 2026-04-10
fix(search): resolve backward AOD rectangle source mapping

### LINKED ISSUE #460: fix(search): correct backward AOD rectangle source mapping
## Summary
Backward-direction bus contexts in `BusContext.from_tree` were building rectangle source coordinates from canonical bus source indices rather than direction-resolved lane endpoints. This caused valid backward movers to be excluded from `pos_to_loc`, so rectangle construction failed and moves were split into singleton steps even when a parallel rectangle was legal.

## Impact
- Valid parallel backward moves can be incorrectly split.
- The planner can emit less parallel move programs than the hardware supports.
- Tuning scoring weights (e.g. very large `w_d`) does not fix the behavior because candidate construction is wrong before scoring/ranking can help.

## Reproduction
1. Run `uv run python debug/steane_physical_impl.py --gif`.
2. Inspect the move sequence around backward word-bus moves (`bus_id=9`, `Direction.BACKWARD`).
3. Observe rows that are lane-group compatible and collision-free getting emitted as separate move statements.

## Root Cause
- `BusContext.from_tree` computed candidate source locations from static bus index sets.
- For backward direction, executable source coordinates differ from forward-side canonical indices.
- `pos_to_loc` therefore missed real movers and `is_valid_rect(...)` returned false on otherwise valid rectangles.

## Proposed Fix
- Build source locations from actual `LaneAddress -> get_endpoints(...)` resolution, which is direction aware.
- Track `src_to_lane` mapping and use it when converting rectangles back to lane sets.
- Keep collision checks source-based using the resolved lane endpoint mapping.

## Validation
- Add regression test: `test_from_tree_backward_word_bus_uses_backward_sources`.
- Confirm search generator tests pass.
- Confirm the Steane debug program merges the two backward rows into one move layer at logical stage.

## Acceptance Criteria
- [ ] Backward bus contexts include real backward source coordinates in `pos_to_loc`.
- [ ] Compatible backward row movers can be emitted in one rectangle moveset.
- [ ] Regression test covers this scenario and passes.
- [ ] Existing search generator tests remain green.

Jason-Ludmir:
## Summary
- fix backward bus rectangle source construction by deriving source locations from direction-aware lane endpoints
- preserve a `src_to_lane` map so rectangle emission reuses valid lane identities instead of reconstructing potentially invalid backward lanes
- add a regression test proving backward word-bus contexts include true backward source coordinates

## Test plan
- [x] `uv run pytest python/tests/search/test_greedy_generator.py`
- [x] regenerate `debug/steane_physical_impl.py --gif` and confirm the previously split backward rows are merged

Closes #460

### DIFF
--- a/python/bloqade/lanes/search/generators/aod_grouping.py
+++ b/python/bloqade/lanes/search/generators/aod_grouping.py
@@ -155,15 +155,7 @@ class BusContext:
                 loc = self.pos_to_loc.get((x, y))
                 if loc is not None:
                     lane = self.src_to_lane.get(loc)
-                    if lane is None:
-                        lane = LaneAddress(
-                            self.move_type,
-                            loc.word_id,
-                            loc.site_id,
-                            self.bus_id,
-                            self.direction,
-                            loc.zone_id,
-                        )
+                    assert lane is not None
                     lanes.append(lane)
         return frozenset(lanes)
 
--- a/python/bloqade/lanes/search/generators/aod_grouping.py
+++ b/python/bloqade/lanes/search/generators/aod_grouping.py
@@ -52,6 +52,9 @@ class BusContext:
     arch_spec: ArchSpec
     pos_to_loc: dict[tuple[float, float], LocationAddress] = field(repr=False)
     collision_srcs: frozenset[LocationAddress] = field(repr=False)
+    src_to_lane: dict[LocationAddress, LaneAddress] = field(
+        repr=False, default_factory=dict
+    )
 
     @classmethod
     def from_tree(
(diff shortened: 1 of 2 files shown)


python/bloqade/lanes/search/generators/aod_grouping.py
```diff
@@ -147,16 +154,17 @@ def rect_to_lanes(self, xs: set[float], ys: set[float]) -> frozenset[LaneAddress
             for y in ys:
                 loc = self.pos_to_loc.get((x, y))
                 if loc is not None:
-                    lanes.append(
-                        LaneAddress(
+                    lane = self.src_to_lane.get(loc)
+                    if lane is None:
+                        lane = LaneAddress(
```
weinbe58: This branch should not be possible, please replace this if statement with an assert.

weinbe58 (approved):
Just address the comment but it should be good to go.

PR #469, opened 2026-04-13
fix(search): add time-aware distance to entropy scoring

### LINKED ISSUE #468: fix(search): Entropy-guided search under-prioritizes time-efficient moves on high-CZ kernels
## Problem
We changed the entropy-guided placement scoring currently to use hop-count distance (lane count) for progress-to-target, rather than approximate move-time distance. This can under-rank candidate moves that are physically time-efficient, and make it so that our solver never converges on a solution.

## Impact
This appears across a broad set of higher-CZ kernels, not just isolated instances. The behavior correlates with elevated search effort and frequent saturation of the per-search expansion budget.

## Observed Symptoms
- Candidate ranking sometimes fails to prioritize moves that significantly improve physical travel cost toward target/CZ-partner placement.
- Search repeatedly spends budget exploring less time-efficient alternatives.
- In benchmark runs, many kernels in higher-CZ ranges reach the `max_expansions` cap.

## Reproduction Context
- Entropy comparison / benchmark flow:
  `uv run python debug/entropy_results/solution_ct_compare.py`
- Tree inspection flow:
  `uv run python debug/viz/visualize_entropy_tree.py ...`

## Expected Behavior
Candidate scoring should reflect physically meaningful progress so time-efficient routing decisions are not systematically deprioritized on high-CZ workloads.

Jason-Ludmir:
## Summary
- Add a time-aware distance term to entropy candidate scoring by blending hop-count shortest-path distance with approximate move-time shortest-path distance.
- Introduce `SearchParams.w_t` (`0.0..1.0`) to control the blend and keep the heuristic tunable across workloads.
- Add bounds constants/validation for `w_t` and tests for min/max enforcement in search-params unit tests.

## Distance Math
For each qubit distance estimate used during candidate scoring, we now compute:

`d_hop = shortest_path(current, target; edge_weight = 1.0)` (lane-count / hops)

`d_time = shortest_path(current, target; edge_weight = lane_duration_us)`

`d_time_norm = d_time / min_lane_duration_us`

`d_blend = (1 - w_t) * d_hop + w_t * d_time_norm`

Where:
- `w_t = 0.0` reproduces the previous hop-only behavior.
- `w_t = 1.0` uses fully time-based distance (normalized into hop-like units).
- normalization by fastest lane duration keeps hop and time terms on comparable scales.

This blended distance feeds per-qubit bus scoring (`delta_d`) and therefore candidate move evaluation/ranking, so moves that improve physical travel cost can be favored even when hop counts are similar.

Closes #468

## Test plan
- [x] `uv run pytest python/tests/search/test_search_params.py`
- [x] pre-commit checks passed during commit (isort, black, ruff, pyright)

### DIFF
--- a/python/bloqade/lanes/search/scoring.py
+++ b/python/bloqade/lanes/search/scoring.py
@@ -7,7 +7,7 @@ Two-level scoring:
 
 from __future__ import annotations
 
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import TYPE_CHECKING
 
 from bloqade.lanes.layout import Direction, LaneAddress, LocationAddress, MoveType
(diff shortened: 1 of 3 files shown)


weinbe58 (approved):
LGTM

PR #480, opened 2026-04-14
feat(search): complete Rust search infrastructure with entropy, cascade, and IDS strategies

lsschmid:
## Summary

Complete Rust implementation of the move synthesis search infrastructure, extending the base A* framework (#436) with multiple search strategies, heuristic improvements, and the time-aware distance blending from #469.

### Search Strategies
- **Entropy-guided search** — port of the Python `EntropyGuidedSearch` with entropy-weighted scoring, reversion on deadlock, and sequential fallback routing
- **Iterative Diving Search (IDS)** — depth-first with heuristic jump-back, inspired by arxiv:2512.13790
- **Cascade** — two-phase strategy: fast inner search (IDS/DFS/Entropy) followed by bounded weighted A* refinement
- **Weighted A\*** — `f = g + weight * h` for bounded suboptimal search
- **DFS** — heuristic depth-first with configurable ordering

### Heuristic Expander Features
- **DeadlockPolicy** — configurable deadlock handling: `Skip` (backtrack), `MoveBlockers` (escape moves for blocking qubits), `AllMoves`
- **FreeRiderPolicy** — opportunistic moves for non-target qubits: `Unblock` or `UnblockOrImprove`
- **2-step lookahead scoring** — evaluates best follow-up move to break ties between candidates
- **Parallel restarts** — rayon-based parallel restarts with seeded score perturbation for diversity

### Time-Aware Distance (port of #469)
- FLAIR constant-jerk motion model ported to Rust (`lane_index.rs`)
- Dijkstra-based time-weighted distance table (conditional on `w_t > 0`)
- Blended distance: `(1 - w_t) * hop_count + w_t * (time / fastest_lane)`
- Default `w_t = 0.05` matching Python's `SearchParams`

### PyO3 Bindings
- String-based strategy selection: `"astar"`, `"ids"`, `"entropy"`, `"cascade-ids"`, `"cascade-entropy"`, etc.
- All parameters exposed: `weight`, `restarts`, `free_riders`, `lookahead`, `deadlock_policy`, `w_t`
- Zone-aware tuple format: `(qubit_id, zone_id, word_id, site_id)`

### Merged from main
- Zone-centric ArchSpec refactor (#437)
- Backward AOD rectangle source mapping fix (#462)
- Time-aware entropy scoring (#469)
- Zone-aware placement helpers (#470)
- RustWrapper base classes (#473)

## Test plan
- [x] `cargo test -p bloqade-lanes-search` — 114 tests pass
- [x] `cargo clippy -p bloqade-lanes-search --all-targets -- -D warnings` — clean
- [x] `cargo fmt --all -- --check` — clean
- [x] `pytest python/tests/heuristics/test_physical_placement.py` — 12 tests pass
- [x] Benchmark smoke test with release build verified


lsschmid:
So, with the rust port I also implemented some alternative searches to Jasons Entropy search. Here also a first rough benchmark on the different version.

exp2_entropy_ablation_small_standard

Jason-Ludmir:
Nice! Can you explain what each of the entropy variants are?

lsschmid:
ah, yes, I added some modifications.

la - this is a lookahead (currently 1 step). This works qubit well to reduce the number of moves but apparently moves more atoms
r28 - this is a restart, so it starts 28 runs of the search with small random variations and takes the best of them. While not beeing deterministic I found that this is usually better than giving more budget to single runs.

I had some more modifications and benchmarks and also run other search algorithms (weighted A*, modified DFS searches, etc.). But I found that your Entropy search works really well and is typicaly a good choice regarding computational cost vs. output quality.

Jason-Ludmir:
Lookahead+restarts are awesome, thank you Ludwig! 

Right now I'm doing some parameter tuning with that AVO loop on the newish version of the code, hopefully using a better parameter set makes a bit of a performance improvement (seeing about ~5% reduction in approx move time).

Also curious how this will do if/when we add having multiple goal nodes to explore in the tree, though this seems fairly similar to your restarts so I don't know if it'll add meaningful improvement over what you implemented here.

crates/bloqade-lanes-search/src/entropy.rs
```diff
@@ -0,0 +1,917 @@
...
+            if occupied.contains(&dst_enc) || visited.contains(&dst_enc) {
+                continue;
+            }
+            if dst_enc == to_enc {
+                let mut full_path = path;
+                full_path.push(lane);
+                return Some(full_path);
+            }
+            visited.insert(dst_enc);
+            let mut new_path = path.clone();
```
weinbe58: **concern**: BFS clones the full path vector at every queue entry. For deep/wide graphs this is O(V×L) memory and O(E×L) time where L is path length. Since the sequential fallback invokes this per unresolved qubit, it could get expensive on large architectures.

Consider using a parent-pointer map (`HashMap<u64, (u64, LaneAddr)>`) and reconstructing the path at the end — O(V) memory, O(L) reconstruction.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
```diff
@@ -216,9 +172,16 @@ impl PyMoveSolver {
         target: Vec<(u32, u32, u32, u32)>,
         blocked: Vec<(u32, u32, u32)>,
         max_expansions: Option<u32>,
-        strategy: PySearchStrategy,
+        strategy: &str,
         top_c: usize,
         max_movesets_per_group: usize,
+        weight: f64,
+        mobility_weight: f64,
```
weinbe58: **concern**: `mobility_weight` is accepted but silently ignored (the Rust `solve()` takes `_mobility_weight`). The docstring says "Reserved for future use" which is honest, but accepting a parameter that does nothing is an API smell — users could waste time tuning a no-op knob.

Consider either removing it until it has an effect, or logging a warning if a non-default value is passed.

weinbe58 (commented):
## PR #480 Review: Rust Search Infrastructure

The core Rust search infrastructure is solid — well-architected `SearchGraph` with arena-based transposition table, clean `Frontier` trait abstraction, and pragmatic cascade strategy. PyO3 bindings have correct GIL handling, good error conversion, and proper input validation. All 20 CI checks pass.

Inline comments below cover the specific findings. Two items that don't fall on diff lines:

**`_MT_MAP` only maps 2 of 3 move types** (`physical_movement.py:~301`): `_MT_MAP = {0: MoveType.SITE, 1: MoveType.WORD}` will `KeyError` if zone bus moves ever appear in search results. The `LaneIndex` currently only generates SiteBus/WordBus lanes so this is safe today, but the expanded strategy set makes it more relevant. Consider adding `2: MoveType.ZONE` defensively.

**No Python-side tests for new parameters**: No tests exercise `weight`, `restarts`, `free_riders`, `lookahead`, `deadlock_policy`, or `w_t` through the Python layer, even as smoke tests. Also no test for invalid strategy strings via the Python API to verify the error path end-to-end.

python/bloqade/lanes/heuristics/physical_movement.py
```diff
@@ -154,7 +154,7 @@ def path_to_target_config(
 class RustPlacementTraversal:
     """Config for the Rust MoveSolver — bypasses ConfigurationTree entirely."""
 
-    strategy: SearchStrategy = SearchStrategy.ASTAR
+    strategy: str = "astar"
```
weinbe58: **concern**: `RustPlacementTraversal` only exposes 4 of the 11 `solve()` parameters (`strategy`, `top_c`, `max_movesets_per_group`, `max_expansions`). The new parameters — `weight`, `restarts`, `free_riders`, `lookahead`, `deadlock_policy`, `w_t` — are not threaded through and always use Rust defaults.

This means the new search features (entropy, cascade, restarts, free riders, deadlock handling) are unreachable from the Python compilation pipeline. If this is intentional for now (Rust-only benchmarking), a comment noting that would help. Otherwise, the dataclass needs corresponding fields.

crates/bloqade-lanes-search/src/entropy.rs
```diff
@@ -0,0 +1,917 @@
...
+
+    let mut graph = SearchGraph::new(root);
+    let root_id = graph.root();
+    let mut entropy_map: HashMap<NodeId, EntropyState> = HashMap::new();
+    let mut current = root_id;
+    let mut nodes_expanded: u32 = 0;
+    let mut max_depth_seen: u32 = 0;
+    let mut found_goals: Vec<NodeId> = Vec::new();
+
+    loop {
```
weinbe58, line 679: **concern**: This `loop` has no hard iteration cap when `max_expansions` is `None`. If the search gets stuck in a reversion cycle without reaching `e_max` at the root, it could spin indefinitely. The sequential fallback acts as a safety valve, but only fires once root entropy hits `e_max` — which depends on the reversion propagation pattern.

Recommend adding a hard iteration limit as a safety net (e.g., `10 * graph_size` or similar), even if callers are expected to always provide a budget in practice.

crates/bloqade-lanes-search/src/frontier.rs
```diff
@@ -202,16 +202,21 @@ impl PartialOrd for PriorityEntry {
...
 }
 
 impl<H> PriorityFrontier<H> {
-    /// Create an A* frontier: `f = g + h`, goal on pop.
-    pub fn astar(heuristic: H) -> Self {
+    /// Create an A* frontier: `f = g + weight * h`, goal on pop.
+    ///
+    /// - `weight = 1.0`: standard A* (optimal with admissible heuristic).
+    /// - `weight > 1.0`: weighted A* (bounded suboptimal, cost ≤ weight × optimal).
+    pub fn astar(heuristic: H, weight: f64) -> Self {
```
weinbe58, line 221: **nit**: The doc comment is good — it documents the weight semantics clearly. However, `goal_on_pop: true` means the search returns the first solution popped from the queue, which is only optimal when `weight == 1.0` with an admissible heuristic. With `weight > 1.0`, the guarantee weakens to bounded suboptimal (cost ≤ weight × optimal). Consider adding a note to the `goal_on_pop` field doc that optimality depends on weight.

crates/bloqade-lanes-search/src/heuristic_expander.rs
```diff
@@ -37,6 +66,16 @@ pub struct HeuristicExpander<'a> {
...
+    /// Policy for deadlock escape.
+    deadlock_policy: DeadlockPolicy,
+    /// Policy for free-rider qubits.
+    free_rider_policy: FreeRiderPolicy,
+    /// Enable 2-step lookahead scoring.
+    lookahead: bool,
+    /// Seed for score perturbation RNG (restart diversity).
+    seed: u64,
+    /// Counter for deadlock occurrences.
+    deadlock_count: AtomicU32,
```
weinbe58: **nit**: `AtomicU32` with `Ordering::Relaxed` works for single-threaded use, but the `Expander` trait doesn't prevent sharing across threads. Since each restart creates its own `HeuristicExpander`, this is safe in practice — but a `Cell<u32>` would more honestly signal the single-threaded intent. Alternatively, a doc comment noting "not intended for sharing across threads" would suffice.

crates/bloqade-lanes-search/src/solve.rs
```diff
@@ -128,9 +157,17 @@ impl MoveSolver {
         strategy: Strategy,
         top_c: usize,
         max_movesets_per_group: usize,
+        weight: f64,
+        _mobility_weight: f64, // reserved — will be used for expander mobility scoring
+        restarts: u32,
+        free_rider_policy: FreeRiderPolicy,
+        lookahead: bool,
+        deadlock_policy: DeadlockPolicy,
+        w_t: f64,
```
weinbe58: **nit**: `solve()` now takes 14 parameters (hence the `#[allow(clippy::too_many_arguments)]`). A `SolveOptions` struct grouping `weight`, `restarts`, `free_rider_policy`, `lookahead`, `deadlock_policy`, and `w_t` would improve ergonomics and make it easier to add future parameters without changing the signature.

python/bloqade/lanes/bytecode/_native.pyi
```diff
@@ -912,20 +902,34 @@ class MoveSolver:
         target: list[tuple[int, int, int, int]],
         blocked: list[tuple[int, int, int]],
         max_expansions: Optional[int] = None,
-        strategy: SearchStrategy = SearchStrategy.ASTAR,
+        strategy: str = "astar",
```
weinbe58: **nit**: `strategy`, `free_riders`, and `deadlock_policy` are typed as `str` but only accept specific values. Using `Literal["astar", "dfs", "bfs", "greedy", "ids", "cascade", "cascade-ids", "cascade-dfs", "cascade-entropy", "entropy"]` (and similar for the others) would give pyright/mypy users compile-time validation instead of runtime `ValueError`.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
```diff
@@ -134,18 +88,25 @@ impl PySolveResult {
...
+
     fn __repr__(&self) -> String {
         let status = match self.inner.status {
             SolveStatus::Solved => "solved",
             SolveStatus::Unsolvable => "unsolvable",
             SolveStatus::BudgetExceeded => "budget_exceeded",
         };
         format!(
-            "SolveResult(status={}, steps={}, cost={}, expanded={})",
+            "SolveResult(status={}, steps={}, cost={}, expanded={}, deadlocks={})",
```
weinbe58: **nit**: The format string produces `SolveResult(status=solved, ...)` without quoting the status. Python convention for `__repr__` would be `SolveResult(status='solved', ...)` so that the output is a valid Python expression. Minor but worth matching the convention.

crates/bloqade-lanes-search/src/heuristic_expander.rs
```diff
@@ -214,6 +331,34 @@ impl Expander for HeuristicExpander<'_> {
...
 
+                // Free-rider augmentation: add non-target qubits sharing
+                // the same bus triplet that unblock or improve.
+                if self.free_rider_policy != FreeRiderPolicy::Off && !moves.is_empty() {
+                    // Reconstruct typed triplet for bus-group filtering.
+                    let mt = if mt_u8 == MoveType::SiteBus as u8 {
+                        MoveType::SiteBus
+                    } else if mt_u8 == MoveType::WordBus as u8 {
+                        MoveType::WordBus
+                    } else {
```
weinbe58: **nit**: The `else { MoveType::ZoneBus }` fallback silently misclassifies if `MoveType` gains new variants. Consider matching exhaustively or using `TryFrom<u8>` to get a compile-time reminder when variants are added.

crates/bloqade-lanes-search/src/lane_index.rs
```diff
@@ -206,6 +225,67 @@ impl LaneIndex {
...
+    let t2 = (-b + discriminant.sqrt()) / (2.0 * a);
+    4.0 * t1 + 2.0 * t2
+}
+
+/// Compute lane duration from waypoints: ramp + sum(segment durations) + ramp.
+fn compute_lane_duration_us(waypoints: &[[f64; 2]]) -> f64 {
+    if waypoints.len() <= 1 {
+        return 0.0;
+    }
+    let ramp = 1.0 / FLAIR_MAX_RAMP_US; // normalized_amp=1.0
```
weinbe58: **nit**: `normalized_amp` is hardcoded to `1.0` here, while the Python version uses the actual FLAIR amplitude delta. This is a reasonable approximation for search heuristics, but a brief doc comment explaining the assumption (e.g., "assumes unit amplitude for search-time cost estimation; exact timing uses FLAIR bytecode values") would help future readers understand the intentional divergence.

python/tests/heuristics/test_physical_placement.py
```diff
@@ -152,7 +151,7 @@ def fake_run_search(self, tree, target, traversal=None):
 
 def test_rust_traversal_default_params():
     t = RustPlacementTraversal()
-    assert t.strategy == SearchStrategy.ASTAR
+    assert t.strategy == "astar"
```
weinbe58, line 156: **nit**: Good test for the default params migration from `SearchStrategy.ASTAR` to `"astar"`. Once `RustPlacementTraversal` exposes the new parameters (`weight`, `restarts`, etc.), consider adding corresponding assertions here, plus a test for invalid strategy strings to verify the error path end-to-end.

crates/bloqade-lanes-search/src/entropy.rs
lsschmid: Fixed — switched to parent-pointer map (`HashMap<u64, (u64, LaneAddr)>`) with path reconstruction. O(V) memory.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
lsschmid: Removed entirely from Rust `solve()`, Python bindings, and type stubs.

python/bloqade/lanes/heuristics/physical_movement.py
lsschmid: Intentional for now — added a docstring noting these are Rust-only until validated via benchmarking. Will thread through in a follow-up.

crates/bloqade-lanes-search/src/entropy.rs
lsschmid: Fixed — added hard iteration limit (`10 * num_locations`) as safety net when `max_expansions` is `None`.

crates/bloqade-lanes-search/src/frontier.rs
lsschmid: Fixed — added doc comment on `goal_on_pop` field noting weight/optimality relationship.

crates/bloqade-lanes-search/src/heuristic_expander.rs
lsschmid: Fixed — switched to `Cell<u32>`.

crates/bloqade-lanes-search/src/solve.rs
lsschmid: Fixed — extracted `SolveOptions` struct with `Default` impl. `solve()` now takes `(initial, target, blocked, max_expansions, &SolveOptions)`.

python/bloqade/lanes/bytecode/_native.pyi
lsschmid: Fixed — `strategy`, `free_riders`, and `deadlock_policy` now use `Literal[...]` types.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
lsschmid: Fixed.

crates/bloqade-lanes-search/src/heuristic_expander.rs
lsschmid: Fixed — exhaustive match with `unreachable!` fallback. Also did the same for `Direction`.

crates/bloqade-lanes-search/src/lane_index.rs
lsschmid: Fixed — added the suggested doc comment.

python/tests/heuristics/test_physical_placement.py
lsschmid: Added a TODO comment. Will add assertions + error-path test when the new params are threaded through to `RustPlacementTraversal`.

lsschmid:
## Search API Refactoring Summary

This batch of commits restructures the Rust search crate into a composable, trait-per-concern architecture matching the Python side. Here's the new structure:

### Core Traits (`traits.rs`)

| Trait | Purpose |
|-------|---------|
| `MoveGenerator` | Produces candidate move sets from a configuration |
| `CandidateScorer` | Ranks candidates (higher = better) for ordering |
| `CostFn` | Computes edge cost for A\* g-score accumulation |
| `Goal` | Decides when the search is complete |
| `Heuristic` | Estimates cost-to-goal (h-function for A\*/greedy) |

### Implementations

**Generators** (`generators/`):

| Type | Description |
|------|-------------|
| `HeuristicGenerator` | Top-C per-qubit filtering, AOD grid construction, deadlock escape (5-15 candidates/expansion) |
| `ExhaustiveGenerator` | Enumerates all valid AOD rectangle move sets |
| `EntropyGenerator` | Entropy-weighted candidate generation (reads per-node entropy from `SearchState`) |
| `GreedyGenerator` | BFS shortest-path per qubit, take first lane, group into AOD grids |

**Scorers** (`scorers/`):

| Type | Description |
|------|-------------|
| `DistanceScorer` | Scores by sum of hop-distance improvements toward targets |
| `EntropyScorer` | `alpha * distance_progress + beta * arrived + gamma * mobility_gain` |

**Goals** (`goals.rs`):

| Type | Description |
|------|-------------|
| `AllAtTarget` | All qubits at exact target locations (standard) |
| `PartialPlacementGoal` | At least N qubits at targets (for multi-phase routing) |

**Heuristics** (`heuristics.rs`):

| Type | Description |
|------|-------------|
| `MaxHopHeuristic` | Max distance over all qubits — admissible, for A\* |
| `SumHopHeuristic` | Sum of distances — not admissible, better ordering for IDS/DFS |

**Cost** (`cost.rs`):

| Type | Description |
|------|-------------|
| `UniformCost` | Every move step costs 1.0 (current default) |

**Observability** (`observer.rs`):

| Type | Description |
|------|-------------|
| `SearchObserver` trait | Callback for monitoring search progress |
| `SearchEvent` enum | `GoalFound`, `NodeExpanded`, `EntropyBump`, `Revert`, `FallbackStart`, `FallbackStep`, `Descend` |
| `NoOpObserver` | Zero-overhead default (discards all events) |

### Shared Context (`context.rs`)

- `SearchContext` — immutable, holds `LaneIndex`, `DistanceTable`, blocked locations, targets
- `SearchState` — mutable per-search state (entropy map for entropy-guided search)
- `MoveCandidate` — output of generators before scoring/cost

### Search Loops

- `frontier::run_search()` — trait-based generic loop for A\*, BFS, DFS, IDS, Greedy
- `entropy::entropy_search()` — single-path DFS with entropy backtracking (uses `Goal` + `SearchContext`)
- `solve::MoveSolver::solve()` — orchestrates everything: strategy dispatch, restarts, cascade two-phase

### Key Changes from Previous Architecture

- `HeuristicExpander` and `ExhaustiveExpander` are gone — logic moved into generators
- Old `Expander` trait internalized (retained for legacy tests only)
- `run_search_v2` renamed to `run_search` (primary API)
- Generators read shared data from `SearchContext` instead of storing references (no lifetime params)
- `FreeRiderPolicy` removed (was always `Off`)
- Two-phase AOD grid construction (`greedy_init` + `merge_clusters`) replaces old rotation-based packing

crates/bloqade-lanes-search/src/aod_grid.rs
```diff
@@ -0,0 +1,449 @@
...
+    ///
+    /// `occupied` is the set of encoded locations currently occupied by atoms.
+    pub(crate) fn new(
+        index: &LaneIndex,
+        mt: MoveType,
+        bus_id: u32,
+        zone_id: u32,
+        dir: Direction,
+        occupied: &HashSet<u64>,
+    ) -> Self {
```
weinbe58, line 46: I want to flag this because the object accepts a `zone_id` it means that we could miss potential optimizations where we have multiple inter-zone moves happening in parallel which is technically possible and defined in the current ArchSpec.

If we're missing some API's in the ArchSpec it would be good to point them out here and add them later.

I'm thinking that we probably need something like `lanes_for(mt, bus_id, dir)` as well. or make `zone_id` an option type.

weinbe58 (changes requested):
Just one comment about the signature of the AOD grouping which might be important for zoned based compilation.

crates/bloqade-lanes-search/src/aod_grid.rs
lsschmid: Fixed in 37c5c44 — zone_id is now optional in BusGridContext and all generators group by (mt, bus_id, dir) without zone, so cross-zone parallel moves on zone buses are supported.

weinbe58 (approved):
LGTM!

PR #490, opened 2026-04-14
feat(benchmarks): add qasm-to-squin conversion utility

### LINKED ISSUE #489: feat(benchmarks): support qasm-to-parallelized-squin benchmark pipeline
## Summary
We want to use OpenQASM benchmark files directly in the benchmarking workflow. We currently have pieces for Cirq→SQUIN conversion and SQUIN block parallelization, but we need a unified, reliable pipeline for QASM inputs.

## Requirements
1. Add a script that converts QASM files into Cirq circuits that are legal to convert downstream (e.g., strips/removes unsupported barriers and terminal measurements when appropriate).
2. Add logic that turns parallelizable `squin.u3`/`squin.cz` regions into `squin.broadcast.*` statements while preserving semantics.
3. Provide an end-to-end solution that takes a QASM file as input and emits a parallelized SQUIN kernel in Python decorator form.

## Dependency Note
Issues #408 and #409 should depend on this issue being resolved (they are blocked until this is complete).

## Acceptance Criteria
- End-to-end CLI/script flow: QASM input → parallelized SQUIN decorator output.
- Conversion handles benchmark QASM files with barriers/terminal measurements.
- Parallelization rewrite preserves statevector-equivalent behavior within numerical tolerance.

Jason-Ludmir:
## Summary
- add `python/benchmarks/utils/qasm_to_squin_kernel.py` to convert OpenQASM inputs into decorator-style SQUIN kernels
- apply the same intended pipeline used for benchmark prep: QASM -> Cirq parse -> optional cleanup (barriers/terminal measurements) -> Cirq parallelization -> SQUIN emission -> block broadcast rewrite (`u3`/`cz` -> `squin.broadcast.*` where valid)
- add package init files for `python/benchmarks/` and `python/benchmarks/utils/`

## Script usage
Run from repo root:

```bash
uv run python python/benchmarks/utils/qasm_to_squin_kernel.py <input.qasm> -o <output.py> --kernel-name <kernel_name>
```

Example:

```bash
uv run python python/benchmarks/utils/qasm_to_squin_kernel.py \
  debug/qasm_bench_small/qpe_n9/qpe_n9_transpiled.qasm \
  -o debug/qasm_bench_small/qpe_n9/qpe_n9_transpiled_squin.py \
  --kernel-name qpe_n9_transpiled
```

## Validation mode
The script includes statevector validation intended for small circuits:

- `--verify-statevector` compares the QASM->Cirq unitary circuit statevector (no extra transforms beyond stripping terminal measurement/barrier parsing issues) against the final generated SQUIN kernel
- `--verify-diagnostics` adds staged SQUIN-vs-SQUIN checks to isolate mismatches
- tolerance controls: `--verify-atol` and `--verify-rtol`

Example with validation:

```bash
uv run python python/benchmarks/utils/qasm_to_squin_kernel.py \
  debug/qasm_bench_small/adder_n4/adder_n4_transpiled.qasm \
  -o debug/qasm_bench_small/adder_n4/adder_n4_transpiled_squin.py \
  --kernel-name adder_n4_transpiled \
  --verify-statevector \
  --verify-diagnostics
```

## Notes
- strips unsupported QASM `barrier` statements before Cirq parsing
- strips terminal measurements before SQUIN conversion
- keeps non-terminal measurement safeguards (errors for unsupported semantics)

Closes #489


python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+    return cirq.optimize_for_target_gateset(circuit, gateset=cirq.CZTargetGateset())
+
+
+def _strip_qasm_barriers(qasm_text: str) -> tuple[str, int]:
+    """Remove OpenQASM barrier statements before Cirq parsing."""
+    kept_lines: list[str] = []
+    dropped = 0
+    for line in qasm_text.splitlines():
+        stripped = line.strip()
+        if stripped.lower().startswith("barrier ") and stripped.endswith(";"):
```
weinbe58: **critical**: `startswith("barrier ")` (note trailing space) will miss the valid OpenQASM form `barrier;` (barrier applied to all qubits with no argument list). This would leave the bare `barrier;` in the QASM text and cause Cirq's parser to fail or produce unexpected results.

Suggestion:
```python
if (stripped.lower().startswith("barrier") 
    and stripped.endswith(";")
    and (len(stripped) == 8 or stripped[7] in (" ", ";"))):
```
Or use a regex: `re.match(r'^barrier(\s.*)?;$', stripped, re.IGNORECASE)`

python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+    q1: int
+
+
+def _format_float(value: float) -> str:
+    return f"{value:.12g}"
+
+
+def _qubit_sort_key(qubit: cirq.Qid) -> tuple[int, str]:
+    if isinstance(qubit, cirq.LineQubit):
+        return (0, str(qubit.x))
```
weinbe58: **critical**: `str(qubit.x)` produces string comparison, so `"10" < "2"` (because `"1" < "2"` lexicographically). For circuits with 10+ qubits, this silently produces wrong qubit index mappings, leading to incorrect SQUIN output.

Fix:
```python
def _qubit_sort_key(qubit: cirq.Qid) -> tuple[int, int | str]:
    if isinstance(qubit, cirq.LineQubit):
        return (0, qubit.x)
    ...
```

The `GridQubit` branch has the same issue — `f"{qubit.row},{qubit.col}"` sorts `"1,0"` before `"10,0"` but after `"0,9"`. Use `(1, qubit.row, qubit.col)` as a tuple key instead.

weinbe58 (commented):
## PR #490 Review: QASM-to-SQUIN Conversion Utility

Neat utility — the pipeline stages are cleanly separated (QASM cleanup → Cirq lowering → SQUIN emission → block parallelization) and the `--verify-statevector` / `--verify-diagnostics` flags are a nice touch for correctness validation.

Two bugs that will produce silently wrong output on 10+ qubit circuits or `barrier;` QASM. Inline comments below.

**Missing tests**: This is a 544-line module with non-trivial gate decomposition and code generation. No accompanying tests in `python/tests/`. Unit tests for `_strip_terminal_measurements`, `_normalize_kernel_name`, `_qubit_sort_key`, `circuit_to_squin_decorator_source`, and `parallelize_squin_blocks` would catch regressions. The verification mode is helpful but requires manual invocation — it won't run in CI.

**`benchmarks` package not in `pyproject.toml`**: The new `python/benchmarks/` directory with `__init__.py` files looks like a package, but it's not registered in `pyproject.toml`. This means `import benchmarks` won't work from an installed package. If it's meant to be a standalone script-only directory, the `__init__.py` files are misleading — consider removing them or adding a comment clarifying the intent. If it should be importable, add it to the packages list.

python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+
+    lines: list[str] = [
+        "from bloqade import squin",
+        "",
+        "@squin.kernel(typeinfer=True, fold=True)",
+        f"def {kernel_name}():",
+        f"    q = squin.qalloc({len(qubits)})",
+    ]
+
+    for op in circuit.all_operations():
```
weinbe58: **concern**: `all_operations()` iterates operations moment-by-moment, but within-moment ordering is not guaranteed by Cirq. After `cirq.optimizers.DropNegligible` + `align_left()` + `parallelize()`, the intra-moment order is arbitrary. This is semantically correct (same-moment ops commute), but the downstream `parallelize_squin_blocks` pass relies on contiguous same-type ops to form broadcast blocks. Non-deterministic intra-moment ordering could fragment blocks across runs, producing different (but equivalent) output for the same input.

python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+def _statevectors_close(
+    lhs: np.ndarray, rhs: np.ndarray, atol: float, rtol: float
+) -> tuple[bool, float]:
+    if lhs.shape != rhs.shape:
+        return False, float("inf")
+
+    lhs_aligned = lhs.copy()
+    rhs_aligned = rhs.copy()
+    overlap = np.vdot(lhs_aligned, rhs_aligned)
+    if np.abs(overlap) > 0:
```
weinbe58: **concern**: The global-phase alignment uses `np.abs(overlap) > 0`, which will almost always be true due to floating-point noise, even for nearly-orthogonal states. When `overlap` is very small (near-zero), `np.angle(overlap)` becomes numerically unstable and the phase correction is meaningless.

Consider a threshold:
```python
if np.abs(overlap) > 1e-12:
    rhs_aligned *= np.exp(-1j * np.angle(overlap))
```

python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+def _simulate_squin_statevector(kernel_path: Path, kernel_name: str) -> np.ndarray:
+    from bloqade.pyqrack.device import StackMemorySimulator
+
+    module_name = f"_qasm_squin_verify_{kernel_path.stem}_{abs(hash(kernel_path))}"
+    spec = importlib.util.spec_from_file_location(module_name, kernel_path)
+    if spec is None or spec.loader is None:
+        raise ValueError(f"Unable to load generated kernel module from {kernel_path}.")
+
+    module = importlib.util.module_from_spec(spec)
+    spec.loader.exec_module(module)
```
weinbe58, line 372: **concern**: `exec_module` on dynamically generated code is inherently risky. The generated code is controlled by this script's output, but the trust chain is: QASM file → float parsing → `_format_float` → emitted Python source → `exec_module`. If a QASM file somehow causes `_format_float` to emit unexpected strings, the generated Python could contain arbitrary code. The risk is low (floats formatted with `:.12g`), but a comment acknowledging this trust boundary would help future readers understand the security model.

python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+    block_ops: list[U3Op | CZOp] = []
+    saw_broadcast = False
+
+    def flush_block() -> None:
+        nonlocal block_kind, block_ops, saw_broadcast
+        if not block_ops:
+            return
+        indent = " " * 4
+        if block_kind == "u3":
+            emitted = _emit_u3_layers(block_ops, indent)  # type: ignore[arg-type]
```
weinbe58, line 303: **nit**: The `# type: ignore[arg-type]` comments suppress type narrowing that could be handled properly. `block_ops` is `list[U3Op | CZOp]` but `_emit_u3_layers` expects `list[U3Op]`. A cleaner approach would be to maintain separate typed lists during accumulation, or use `cast()` to make the narrowing explicit:
```python
from typing import cast
emitted = _emit_u3_layers(cast(list[U3Op], block_ops), indent)
```

python/benchmarks/utils/qasm_to_squin_kernel.py
```diff
@@ -0,0 +1,544 @@
...
+                    )
+            measurement_count += 1
+
+        if kept_ops:
+            kept_moments.append(cirq.Moment(kept_ops))
+
+    return cirq.Circuit(kept_moments), measurement_count
+
+
+def _to_cz_native_cirq(circuit: cirq.Circuit) -> cirq.Circuit:
```
weinbe58, line 99: **nit**: `_to_cz_native_cirq` is defined at module level but only used inside the `--verify-diagnostics` branch of `main()`. Consider either moving it into the diagnostics section or adding a brief doc note that it's diagnostics-only, so readers don't wonder why the main pipeline doesn't use it.

weinbe58:
**suggestion**: Cirq has built-in QASM2 import support via `cirq.Circuit.from_qasm()` which could simplify the parsing pipeline. It already handles barrier statements, qubit ordering, and basic gate decomposition — many of the edge cases being worked around here manually. Worth checking if `cirq.Circuit.from_qasm(qasm_text)` covers enough of the pipeline to replace the manual QASM text preprocessing and reduce the surface area for bugs like the barrier/qubit-sort issues noted above.

python/benchmarks/utils/qasm_to_squin_kernel.py
Jason-Ludmir: Fixed. Barrier handling now covers both `barrier;` and `barrier q[...]...;` via regex-based matching in the fallback parser path, and primary parsing now uses `cirq.Circuit.from_qasm()` first. Added regression tests for both barrier forms.

python/benchmarks/utils/qasm_to_squin_kernel.py
Jason-Ludmir: Fixed. `_qubit_sort_key` now uses numeric tuple keys (`LineQubit.x`, `GridQubit.row/col`) instead of string conversion, eliminating lexicographic misordering for 10+ qubits. Added regression test validating numeric ordering.

python/benchmarks/utils/qasm_to_squin_kernel.py
Jason-Ludmir: Addressed by deterministic per-moment op ordering before emission, plus canonical CZ index ordering, so generated SQUIN source is stable across equivalent intra-moment op orderings. Added deterministic output regression test.

python/benchmarks/utils/qasm_to_squin_kernel.py
Jason-Ludmir: Addressed by applying phase alignment only when `|overlap| > 1e-12` to avoid unstable angle correction in near-orthogonal cases. Added regression test for tiny-overlap behavior.

Jason-Ludmir:
Addressed your suggestion about `cirq.Circuit.from_qasm`: the loader now prefers Cirq's built-in QASM import path first, with fallback to `cirq.contrib.qasm_import.circuit_from_qasm` only when needed. Barrier stripping is retained only in the fallback path (now regex-based for `barrier;` and explicit forms), which reduces manual preprocessing surface area while preserving compatibility.

weinbe58 (approved):
LGTM

PR #521, opened 2026-04-16
fix(search): make Rust candidate ordering deterministic

### LINKED ISSUE #520: fix(search): investigate nondeterministic Rust solver outcomes
## Summary
Benchmark runs show nondeterministic pass/fail behavior for Rust search strategies (notably `rust_ids`) on the same inputs and configuration.

## Reproduction
1. Run `just benchmark-physical` repeatedly with the same code and matrix.
2. Observe occasional flip-flops in `success` for some `(case_id, strategy_id)` pairs (e.g. `steane_physical_35/rust_ids`).
3. Observe metric drift (`nodes_explored`, `move_count_events`) beyond expected timing variance.

## Why this is a problem
- Benchmark comparisons become unstable and hard to trust.
- CI/PR checks can fail intermittently on unchanged code.
- Strategy quality/regression analysis is noisy.

## Suspected root causes
- Non-deterministic iteration/order effects in Rust search candidate generation.
- Hash-map iteration ordering and tie-breaking sensitivity under expansion budgets.

## Scope of follow-up fix
- Make Rust search traversal/candidate ordering deterministic for identical inputs.
- Ensure seeded/restart behavior is explicit and reproducible.
- Add regression coverage to assert repeatable outcomes across repeated runs.

Jason-Ludmir:
## Summary
- remove hash-order dependence in Rust candidate grouping/selection for search by introducing deterministic ordering and explicit tie-breakers
- apply deterministic ordering in both `entropy.rs` and `generators/heuristic.rs` so equal-score cases explore candidates consistently across runs
- add targeted regression tests that assert stable candidate ordering and repeatable behavior for identical inputs

## Why this fixes nondeterminism
The prior implementation relied on `HashMap` iteration and score-only sorts at key stages. For equal/near-equal candidates, process-level hash randomization could change exploration order, which in turn changed outcomes under expansion budgets. This patch enforces a stable ordering via deterministic maps and tie-break sorting keys, making repeated runs reproducible.

## Test plan
- pre-commit hooks passed on commit (`cargo fmt`, `cargo clippy`, `pyright-check`)
- added/updated Rust tests in the modified files to validate deterministic ordering

Closes #520

### DIFF
--- a/crates/bloqade-lanes-search/src/generators/heuristic.rs
+++ b/crates/bloqade-lanes-search/src/generators/heuristic.rs
@@ -20,6 +20,7 @@ use crate::context::{MoveCandidate, SearchContext, SearchState};
 use crate::graph::{MoveSet, NodeId};
 use crate::heuristic::DistanceTable;
 use crate::lane_index::LaneIndex;
+use crate::ordering::{TripletKey, cmp_moveset_config_tiebreak, cmp_triplet_entry_tiebreak};
 use crate::traits::MoveGenerator;
 
 /// Policy for handling deadlocks (no improving moves available).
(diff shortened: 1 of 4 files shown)


crates/bloqade-lanes-search/src/generators/heuristic.rs
```diff
@@ -43,6 +44,21 @@ struct ScoredTriple {
...
+        .then_with(|| a.1.dst_encoded.cmp(&b.1.dst_encoded))
+        .then_with(|| a.1.qubit_id.cmp(&b.1.qubit_id))
+}
+
+fn cmp_candidates(a: &(i32, MoveSet, Config), b: &(i32, MoveSet, Config)) -> Ordering {
+    b.0.cmp(&a.0)
+        .then_with(|| a.1.encoded_lanes().cmp(b.1.encoded_lanes()))
+        .then_with(|| a.2.as_entries().cmp(b.2.as_entries()))
+}
+
```
weinbe58, line 65: to make the code more DRY and share the implementation from the other module to prevent drift in the convention

weinbe58 (changes requested):

crates/bloqade-lanes-search/src/generators/heuristic.rs
Jason-Ludmir: Refactored to share ordering logic across modules to prevent drift. I added a shared `ordering.rs` helper for triplet and candidate tie-break conventions, then wired both `entropy.rs` and `generators/heuristic.rs` to use those helpers (commit `029b661`).

weinbe58 (approved):
LGTM

PR #536, opened 2026-04-21
feat(search): Rust-native target generator plugin system

lsschmid:
## Summary

Rust-side complement to the Python target generator plugin from #533. Where #533 adds `TargetGeneratorABC` in Python (generate → validate → solve loop in Python with per-candidate `solver.solve()` calls), this PR moves the entire loop into Rust:

- **Python (#533):** Each candidate requires a Python → Rust → Python round-trip (GIL released during solve, but re-acquired between candidates).
- **Rust (this PR):** `solve_with_generator()` runs the full generate → validate → solve loop in Rust with a single FFI call and the GIL released throughout.

Both systems coexist — Python generators use `TargetGeneratorABC`, Rust generators use the `TargetGenerator` trait. A `DefaultTargetGenerator` in Rust replicates the Python default (CZ blockade partner lookup).

### New API

**Rust:** `TargetGenerator` trait, `DefaultTargetGenerator`, `validate_candidate()`, `MoveSolver::solve_with_generator()`, `MoveSolver::generate_candidates()`, `MultiSolveResult` with per-candidate `CandidateAttempt` debug info.

**Python (via PyO3):** `DefaultTargetGenerator`, `MultiSolveResult`, `MoveSolver.solve_with_generator()`, `MoveSolver.generate_candidates()`.

## Test plan

- [x] 156 Rust tests pass (8 new: 5 unit + 3 integration)
- [x] 9 new Python tests pass
- [x] `cargo fmt`, `cargo check`, `cargo clippy` clean


crates/bloqade-lanes-bytecode-python/src/search_python.rs
```diff
@@ -193,116 +194,392 @@ impl PyMoveSolver {
...
+    ///     w_t: Time-distance blend (default 0.05).
+    ///
+    /// Returns:
+    ///     MultiSolveResult with per-candidate debug info.
+    #[pyo3(signature = (initial, blocked, controls, targets, generator=None, max_expansions=None, strategy="astar", top_c=3, max_movesets_per_group=3, weight=1.0, restarts=1, lookahead=false, deadlock_policy="skip", w_t=0.05))]
+    #[allow(clippy::too_many_arguments)]
+    fn solve_with_generator(
+        &self,
+        py: Python<'_>,
+        initial: Vec<(u32, u32, u32, u32)>,
```
weinbe58: Make sure to use the relevant structures for readability. For example I assume here that this is supposed to encode the location addresses for each qubit id. I don't think the LocationAddress give any overhead in terms of performance in rust.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
```diff
@@ -193,116 +194,392 @@ impl PyMoveSolver {
...
+        "cascade" | "cascade-ids" => Strategy::Cascade {
+            inner: InnerStrategy::Ids,
+        },
+        "cascade-dfs" => Strategy::Cascade {
+            inner: InnerStrategy::Dfs,
+        },
+        "cascade-entropy" => Strategy::Cascade {
+            inner: InnerStrategy::Entropy,
+        },
+        "entropy" => Strategy::Entropy,
```
weinbe58: Not blocking we can fix this later but I would like to implement enums for the python interfaces as well

crates/bloqade-lanes-bytecode-python/src/search_python.rs
weinbe58: If the issue is the difference between the rust and python objects the new wrappers should have a way of extracting the rust objects at the callsite of the py03 API

crates/bloqade-lanes-bytecode-python/src/search_python.rs
```diff
@@ -193,116 +194,392 @@ impl PyMoveSolver {
...
+
+#[allow(clippy::too_many_arguments)]
+fn build_solve_options(
+    strategy: &str,
+    top_c: usize,
+    max_movesets_per_group: usize,
+    weight: f64,
+    restarts: u32,
+    lookahead: bool,
+    deadlock_policy: &str,
```
weinbe58: To reduce the number of arguments creating a struct with the fields could be helpful. You can even make it a python object so it is shared across the rust-python boundary.

weinbe58 (changes requested):
overall looks good I think we need some work on building some abstractions for carring around the solver parameters as well as making sure to use the abstractions at the boundary to make it clear what data is being passed around. 

I know currently it is a bit awkward because there are python objects are thin wrappers of the rust objects added in https://github.com/QuEraComputing/bloqade-lanes/pull/473

lsschmid (commented):
Thanks for the review! I've addressed the immediate feedback (enums, SolveOptions, typed LocationAddress on the new methods). 

For the broader refactoring — making the wrapper types the consistent boundary representation across all solver methods (including migrating the existing `solve()` API away from raw tuples) — I'll create a separate follow-up PR to keep this one focused.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
lsschmid: Addressed in 7349afb — `solve()`, `solve_with_generator()`, and `generate_candidates()` now accept `dict[int, LocationAddress]` and `list[LocationAddress]` instead of raw tuples. `goal_config` also returns `dict[int, LocationAddress]`.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
lsschmid: Addressed in 74bceba — added `SearchStrategy` and `DeadlockPolicy` enum pyclasses.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
lsschmid: Addressed together with comment 6 — the PyO3 methods now accept `PyLocationAddr` directly and access `.inner` at the boundary.

crates/bloqade-lanes-bytecode-python/src/search_python.rs
lsschmid: Addressed in 74bceba — added `SolveOptions` pyclass that bundles all tuning parameters. All solver methods now accept `options: SolveOptions | None` instead of individual params.

lsschmid (commented):
All feedback has been addressed in this PR:

- **Enums** (comment 7): `SearchStrategy` + `DeadlockPolicy` pyclasses (74bceba)
- **SolveOptions** (comment 9): Bundles all tuning params into a single struct (74bceba)  
- **Typed LocationAddress** (comments 6+8): `solve()`, `solve_with_generator()`, `generate_candidates()` and `goal_config` all use `LocationAddress` instead of raw tuples (7349afb)
- Dead string-parsing helpers removed

weinbe58 (approved):
LGTM

Issue #544, opened 2026-04-22
Share opcode names between SST text parser and PyO3 op_name() binding


weinbe58:
## Summary

Right now the canonical opcode name strings live in two places that can drift:

1. The SST text parser's forward/reverse match arms in `crates/bloqade-lanes-bytecode-core/src/bytecode/text.rs` (string literals like `\"const_float\"`, `\"move\"`, `\"return\"`).
2. The Python `Instruction.op_name()` accessor in `crates/bloqade-lanes-bytecode-python/src/instruction_python.rs` (added in #541 — bytecode accessors).

A drift between these two already happened once (caught in #541 review — `op_name()` was returning `\"move_\"` / `\"return_\"` while the parser used `\"move\"` / `\"return\"`). A single source of truth would prevent future drift.

### Core crate: add name constants + method on `Instruction`

In `crates/bloqade-lanes-bytecode-core/src/bytecode/instruction.rs`:

```rust
pub const NAME_CONST_FLOAT: &str = \"const_float\";
pub const NAME_CONST_INT: &str = \"const_int\";
pub const NAME_MOVE: &str = \"move\";
pub const NAME_RETURN: &str = \"return\";
// ... one const per opcode (22 total)

impl Instruction {
    pub fn name(&self) -> &'static str {
        match self {
            Instruction::Cpu(CpuInstruction::ConstFloat(_)) => NAME_CONST_FLOAT,
            Instruction::Cpu(CpuInstruction::ConstInt(_)) => NAME_CONST_INT,
            // ...
        }
    }
}
```

The compiler enforces exhaustiveness on the enum match — any future opcode addition is caught at compile time.

### Parser (`text.rs`): reference the consts

- **Reverse direction** — pattern-match on consts (Rust's `match` accepts `const &str` patterns):

  ```rust
  match op_str {
      NAME_CONST_FLOAT => { /* parse operand */ }
      NAME_MOVE => { /* parse arity */ }
      NAME_RETURN => Ok(Instruction::Cpu(CpuInstruction::Return)),
      // ...
  }
  ```

- **Forward direction** — use `instr.name()` in `format_instruction` instead of duplicated literal RHSes:

  ```rust
  format!(\"{} {}\", instr.name(), format_float(f))
  ```

### PyO3 binding: delegate

In `crates/bloqade-lanes-bytecode-python/src/instruction_python.rs`:

```rust
fn op_name(&self) -> &'static str {
    self.inner.name()
}
```

[...]

PR #541, opened 2026-04-22
feat(bytecode): add Instruction operand accessors (PyO3)

weinbe58:
## Summary

Adds per-opcode operand accessors to the Rust-side `Instruction` PyO3 binding so the Python decoder (tracked separately in `feat/stack-move-impl`) can introspect instructions. Previously only `Instruction.opcode: int` was exposed.

This is **Phase 0** of the stack_move dialect design (see `docs/superpowers/specs/2026-04-21-stack-move-dialect-design.md` and `docs/superpowers/plans/2026-04-21-stack-move-dialect.md` on the base branch `spec/stack-move-dialect`).

### New accessors on `Instruction`

| Method | Applies on | Returns |
|---|---|---|
| `op_name()` | every opcode | `str` (snake_case factory-method name, e.g. `"const_float"`, `"move_"`, `"return_"`) |
| `arity()` | `initial_fill`, `fill`, `move_`, `local_r`, `local_rz`, `measure` | `int` |
| `float_value()` | `const_float` | `float` |
| `int_value()` | `const_int` | `int` |
| `location_address()` | `const_loc` | `LocationAddress` (PyO3) |
| `lane_address()` | `const_lane` | `LaneAddress` (PyO3) |
| `zone_address()` | `const_zone` | `ZoneAddress` (PyO3) |
| `type_tag()` / `dim0()` / `dim1()` | `new_array` | `int` / `int` / `int` |
| `ndims()` | `get_item` | `int` |

Each accessor raises `RuntimeError` when called on an opcode it doesn't apply to. All accessors are also documented in the `.pyi` type stub.

### Files

- `crates/bloqade-lanes-bytecode-python/src/instruction_python.rs` — added 11 `#[pymethods]` accessors with exhaustive match over `rs::Instruction` variants.
- `python/bloqade/lanes/bytecode/_native.pyi` — stub docstrings for all accessors.
- `python/tests/bytecode/test_bytecode.py` — new `TestInstructionAccessors` class exhaustively covering all 24 opcodes + error-path tests.

### Commits

- `afe86cb` — `op_name()` dispatcher
- `3be82d9` — `arity()`
- `25ccf35` — `float_value()` / `int_value()`
- `f7e95dc` — address accessors (location/lane/zone)
- `a5911af` — array accessors (type_tag/dim0/dim1/ndims)

## Test plan

- [x] `uv run pytest python/tests/bytecode/` — 153 passed
[...]

Issue #248, opened 2026-03-13
Refactor move dialect to match bytecode execution semantics


weinbe58:
## Summary

Refactor the `move` dialect (`python/bloqade/lanes/dialects/move.py`) so its IR statements match the execution semantics of the Rust bytecode. The key change is that addresses (`LocationAddress`, `LaneAddress`, `ZoneAddress`) should become **SSA values** (produced by `Const*` statements) rather than **statement attributes**, mirroring how the bytecode pushes address constants onto the stack before operations consume them.

**Depends on**: #246 (Rust/Python type unification — the `Const*` statements will wrap the unified Rust-backed address objects)

**Assumption**: Functions contain a single control-flow block (consistent with current bytecode).

## Motivation

Currently, code generation from the move dialect to bytecode requires extracting address attributes from each statement and emitting the corresponding `ConstLoc`/`ConstLane`/`ConstZone` bytecode instructions. By making the SSA-IR mirror the bytecode's structure (addresses as values, explicit arity), codegen becomes mechanical.

### Current dialect (addresses as attributes)

```
%state = move.Load()
%state1 = move.Fill(%state) [location_addresses=(LocationAddress(0,0), LocationAddress(0,1))]
%state2 = move.Move(%state1) [lanes=(LaneAddress(...), LaneAddress(...))]
%state3 = move.CZ(%state2) [zone_address=ZoneAddress(0)]
move.Store(%state3)
```

### Proposed dialect (addresses as SSA values)

```
%state = move.Load()
%loc0 = move.ConstLoc() [value=LocationAddress(0,0)]
%loc1 = move.ConstLoc() [value=LocationAddress(0,1)]
%state1 = move.Fill(%state, %loc0, %loc1)
%lane0 = move.ConstLane() [value=LaneAddress(...)]
%lane1 = move.ConstLane() [value=LaneAddress(...)]
%state2 = move.Move(%state1, %lane0, %lane1)
%zone0 = move.ConstZone() [value=ZoneAddress(0)]
%state3 = move.CZ(%state2, %zone0)
move.Store(%state3)
```

### New IR types

```python
LocationAddressType = ...  # SSA type for LocationAddress values
LaneAddressType = ...      # SSA type for LaneAddress values
ZoneAddressType = ...      # SSA type for ZoneAddress values
```

### New statements: address constants

These wrap the Rust-backed address objects (from #246) as compile-time attributes and produce typed SSA values:

```python
[...]

PR #545, opened 2026-04-23
feat(stack_move): dialect, decoder, lower_stack_move, measure_lower

weinbe58:
## Summary

Implements the full Python-side `stack_move` pipeline: a 1:1 SSA dialect mirroring the bytecode, a `BytecodeDecoder` that turns a Rust `Program` into a `stack_move` `ir.Method`, a `RewriteStackMoveToMove` rewrite that lowers `stack_move` into the existing multi-dialect target IR (`move`, `ilist`, `py.constant`, `py.indexing`, `annotate`, `func`), extensions to `AtomAnalysis` (the `MeasureFuture` lattice element now carries a `measurement_count` ordinal), and a `measure_lower` rewrite that collapses `move.Measure` into `move.EndMeasure` by consulting the analysis frame.

## Relationship to #248

This PR is the foundational step toward closing [#248 — *Refactor move dialect to match bytecode execution semantics*](https://github.com/QuEraComputing/bloqade-lanes/issues/248). It adds `stack_move` as the canonical bytecode-shaped SSA representation, but it does **not** fully migrate the existing `move` dialect or its production flows — those coexist with `stack_move` on the new additive path.

We are **not** auto-closing #248 with this PR. The remaining work (auditing every `move` producer, deciding whether `move` becomes a thin target-dialect shim or is retired entirely in favour of `stack_move` + a few stateful ops, updating downstream passes) is tracked in a dedicated follow-up tracking issue: **#551**.

## Highlights

1. **`stack_move` dialect** — 1:1 SSA image of the bytecode (12 statement families, linear-IR style with explicit `Pop`/`Dup`/`Swap`). Bytecode opcodes that overlap with `kirin.prelude.basic` (`return` / `halt`) are intentionally *not* duplicated — the decoder emits `func.Return` / `func.ConstantNone` + `func.Return` directly.
2. **`BytecodeDecoder`** — syntactic `Program` → `stack_move` `ir.Method`, with a `StackMachineFrame` mirroring Kirin's `lowering.Frame` idiom. `load_program` runs a `TypeInfer` pass before returning.
3. **`RewriteStackMoveToMove`** — in-place per-statement rewrite. Requires an `ArchSpec` (consumed by the `AwaitMeasure` lowering, which expands into one `move.GetFutureResult` per `(zone, location)` from `arch_spec.yield_zone_locations`). `rewrite_Block` owns the block's state boundary: inserts `move.Load` at the start and a closing `move.Store(state)` before the terminator.
4. **New `move.Measure` statement** — `StatefulStatement` subclass with `zone_addresses` compile-time attribute + two results (state + future). Produced by `stack_move2move`, consumed by `measure_lower`.
5. **`AtomAnalysis` extension** — `MeasureFuture` lattice element gains a `measurement_count: int` ordinal populated from the running `final_measurement_count` counter on the interpreter after incrementing.
6. **`measure_lower`** — reads each `move.Measure`'s `future` SSA from a stashed `ForwardFrame` and rewrites to `move.EndMeasure` when preconditions hold (one zone observed, `measurement_count == 1`). Gives up silently on precondition failure — validation is a dedicated-pass concern, not a rewrite-rule concern.

## End-to-end tests

Two tests in `python/tests/test_stack_move_e2e.py`:

- **Minimal**: `fill` → `measure` → `await` → `return`. Asserts the full expected block structure after both rewrites (`move.Load`, `move.Fill`, `move.EndMeasure`, `move.GetFutureResult × N`, `ilist.New`, `move.Store`, `func.Return`).
- **Realistic CZ-sandwich kernel**: 4 atoms filled at zone 0 words 0/2/4/6; atom 0 hopped 0→1→3 via a 2-lane `move_(2)` (bus 0 forward 0→1, then bus 9 forward 1→3 — `apply_moves` runs lanes sequentially); `CZ` on zone 0; atom 0 moved back 3→0 via the reverse 2-lane `move_(2)`; atoms 2 and 3 moved 4→1 and 6→3 via a single 2-lane `move_(2)` on bus 5 forward; `CZ`; atoms 2 and 3 moved back 1→4 and 3→6 via bus 5 backward; `measure(1)` on zone 0; `await_measure`; four per-index `get_item(1)` lookups at the original fill word_ids (0, 2, 4, 6), with `dup`/`swap` between lookups to keep the measurement-array reference on top of the stack across extractions; `new_array(type_tag=Int, dim0=4, dim1=0)`; `return_`. The assertion builds the expected multi-dialect SSA block by hand (state-threading dependencies wired via walrus assignments) and compares it structurally via `bloqade.test_utils.assert_nodes`.

## Test plan

- [x] `uv run pytest python/tests/` — **1030 passed, 4 skipped**.
- [x] `cargo clippy -p bloqade-lanes-bytecode-core -p bloqade-lanes-bytecode-cli --all-targets -- -D warnings` — clean (CI-scoped crates).
- [x] Black + isort + ruff — clean.

## Design doc

`docs/superpowers/specs/2026-04-21-stack-move-dialect-design.md` — updated on this branch to reflect the shipped implementation (signatures, rewrite contract, `AwaitMeasure` lowering semantics, `measure_lower` silent-give-up contract, alignment with `kirin.prelude.basic`).

## Dependencies

- Base: `spec/stack-move-dialect` (contains #541's PyO3 operand accessors + design docs + plan).

## Follow-ups (not blocking)

- #551 — tracking the `move` dialect migration (this is the follow-up to #248).
- #544 — share opcode name strings between SST text parser and PyO3 `op_name()` (prevents future drift).
- #547 — add `TAG_MEASUREMENT_RESULT` to the Rust value tags so `new_array` can accurately tag a new array of `MeasurementResult` elements without an `Int`-tag placeholder.


weinbe58:
Addressed the five review items you left (in `docs/superpowers/feedback/2026-04-21-stack-move-dialect.md` plus the new `StackMachineFrame` suggestion):

| # | Commit | Change |
|---|---|---|
| 1 | `7439c61` | `stack_move` address types now come from `bloqade.lanes.layout.encoding` (Python wrappers), not `bloqade.lanes.bytecode` (Rust-native). Drops the G1 `_lift_{location,lane,zone}_addresses` boundary helpers + `cast()` workarounds. |
| 2 | `56eafd9` | `stack_move.Return` now takes `value: ir.SSAValue`. Decoder pops one; rewrite threads it through to `func.Return` (no more `func.ConstantNone()` workaround). |
| 3 | `3fbe756` | Generic types on Pop/Dup/Swap: `Dup(value: T) -> T` for identity, `Swap(in_top: Top, in_bot: Bot) -> (Bot, Top)` for the permutation invariant, `Pop(value: Any)`. Used `types.TypeVar` (Kirin's API, matches in-tree convention — `types.Generic` is for parameterised generic types, not type variables). |
| 4 | `5c3f40d` | Gate statement fields renamed `phi → axis_angle`, `theta → rotation_angle`, matching the `move` dialect. Rewrite's kwarg forwarding is now a clean pass-through. |
| 5 | `c30a9b6` | Extracted `StackMachineFrame` from `BytecodeDecoder`. Frame owns `block` + `stack` + `push_value`/`pop_value`/`peek_value`/`swap_values`/`append`. Decoder delegates — pure dispatch. Mirrors `kirin.lowering.Frame`. |

All 270 tests still green across `dialects/`, `bytecode/`, `rewrite/`, `analysis/atom/`, and the e2e test.

weinbe58:
Addressed the four additional review items:

| # | Commit | Change |
|---|---|---|
| 1 | `d9cfa2e` | `RewriteStackMoveToMove` now uses `@functools.singledispatchmethod` for per-statement-type dispatch instead of `getattr(self, f"_rewrite_{TypeName}")` string-based lookup. Matches the `place2move.py::RewriteGates.stmts_to_insert` pattern. 24 handlers renamed to `_` + decorated with `@_rewrite.register(stack_move.X)`. |
| 2 & 3 | `8245165` | Terminology alignment: `bytecode/lowering.py` → `bytecode/decode.py`, `LoweringError` → `DecodingError`, `rewrite/lower_stack_move.py` → `rewrite/stack_move2move.py`, `LowerStackMove` → `RewriteStackMoveToMove`. Test files renamed correspondingly. `git mv` used to preserve history (90–98% similarity confirmed). |
| 4 | `5ed77f2` | `Array`/`ArrayType` are now Kirin generics parameterised by `(ElemType, Dim0Type, Dim1Type)`. `stack_move.NewArray` is `init=False` with a custom `__init__` that derives a fully-parameterised `ArrayType[TYPE_TAG[type_tag], Literal(dim0), Literal(dim1)]` as its result type. `stack_move.GetItem` is type-parametric over `ElemType` so its result matches the array's element type. `TYPE_TAG` maps the 9 bytecode type-tag bytes (from `value.rs`) to their Kirin element types. Decoder passes `values=()` — bytecode's `new_array` doesn't pop. |

All 270 tests still green. PR updated via `--force-with-lease`.

weinbe58:
Restructured `StackMachineFrame` per follow-up spec (`5f883fd`):

- `current_region: ir.Region` + `current_block: ir.Block` fields (mirrors `kirin.lowering.Frame`), both initialised in `__post_init__` with the region owning the block.
- `push(stmt: T) -> T` replaces `append`, returning the statement for chaining.
- Call-site idioms adopted across 21 `_visit_*` handlers:
  - **Single-result**: `result = self.frame.push(stmt).expect_one_result()` (11 handlers — constants, `Dup`, `Measure`, `NewArray`, `GetItem`, `SetDetector`, `SetObservable`)
  - **Zero-result side-effect**: bare `self.frame.push(stmt)` (12 handlers — `Return`, `Pop`, atom ops, gates, `AwaitMeasure`, `Halt`)
  - **Multi-result**: `stmt = self.frame.push(Swap(...))` then `stmt.out_bot` / `stmt.out_top` (1 handler — `Swap`)
- `_finalize` now uses `self.frame.current_region` directly instead of rebuilding a Region from a block.

Used Kirin v0.22's `expect_one_result()` method name (verified against `release-0-22` branch — the `.expect_one()` shorthand in the spec is `expect_one_result()` in v0.22). Pure API-surface refactor, 270/270 tests still pass.

weinbe58:
Refinement to the last refactor (`97ea98e`): `StackMachineFrame.push` now also pushes the statement's SSA results onto the virtual stack, in **reverse declaration order**.

- For single-result statements (12 handlers), the decoder is now a one-liner: `self.frame.push(stack_move.ConstFloat(value=...))` — the result auto-pushes.
- For `Swap` (the only multi-result stmt), reversed iteration over `(out_top, out_bot)` lands `out_top` on top after `out_bot` — matching the previous explicit behaviour.
- Zero-result stmts (Fill/Move/gates/etc.) are unaffected — auto-push is a no-op.
- `AwaitMeasure` keeps its explicit `push_value(future)` because the bytecode re-pushes the future; not a statement-result case.

The reverse-order convention mirrors how arguments are declared top-first and popped top-first — keeps the dialect declaration order = stack-position semantics symmetric.

weinbe58:
Moved stack-underflow semantics into the frame (`492248a`):

- `StackMachineFrame.pop_value()` / `pop_n(n)` / `peek_value()` / `swap_values()` now all raise a frame-level `StackUnderflowError` on underflow.
- `BytecodeDecoder._visit` catches and rewraps into `DecodingError` with instruction context (idx + opcode name).
- `_pop_or_raise` / `_pop_n` removed from `BytecodeDecoder` — 19 handler call sites simplified (`self._pop_n(idx, instr, arity)` → `self.frame.pop_n(arity)`).
- Handlers no longer thread `idx, instr` through stack ops; they still receive them for non-stack uses (instruction operand accessors like `instr.arity()`).

Existing underflow test still catches `DecodingError` at the public API — no test changes needed. 270/270 still green.

weinbe58:
Fixed the `AwaitMeasure` semantics (`09c566f`):

- `stack_move.AwaitMeasure` now correctly types its result as `ArrayType[MeasurementResultType, Any, Any]` — matches the bytecode spec ("pops a measure future and pushes an array reference containing the measurement results"). The future is a linear type, consumed not re-pushed.
- Decoder's `_visit_await_measure` simplifies — `frame.push` auto-pushes the new array-ref result. Dropped the explicit `push_value(future)` re-push and the stale comment claiming it was pure synchronisation.
- Rewrite emits an `ilist.New(values=())` placeholder for the array ref so downstream consumers have a valid target-dialect SSA value. Proper per-location `GetFutureResult` expansion is a known v1 stub — needs dialect-level future-to-array bridge work.

Follow-up: #547 tracks the missing Rust `TAG_MEASUREMENT_RESULT` (0x9) tag for the measurement-result array element type. The Python-side `AwaitMeasure` result type is statically known so doesn't depend on the Rust tag, but any direct `new_array(type_tag=...)` call wanting to create an array of `MeasurementResult` would.

270/270 tests still green.

weinbe58:
Tightened `SetDetector` / `SetObservable` array argument types (`978220b`):

```python
array: ir.SSAValue = info.argument(
    type=ArrayType[MeasurementResultType, types.Any, types.Literal(0)]
)
```

- Element type: `MeasurementResultType` (the AwaitMeasure output element type).
- dim0: `Any` (any length).
- dim1: `Literal(0)` (1-D per the bytecode's convention where `dim1=0` means 1-D).

Validation of the constraint is the job of type-inference passes — the decoder doesn't check, nor do construction sites fail at runtime. 270/270 tests still green.

weinbe58:
Self-consistency audit between the Rust validator and the Python docs/dialect turned up two semantic bugs and four incomplete docstrings. All fixed:

| Commit | Fix |
|---|---|
| `3346e3b` | `measure(arity)` pops **zones** (not locations) and pushes **arity** measure futures (not 1). `stack_move.Measure` is now `init=False` with a custom `__init__` producing `arity` results (mirrors `NewArray`'s pattern). Decoder + rewrite + all relevant tests + e2e test all updated — the e2e test's `const_loc` before `measure(1)` is now `const_zone(0)`, which was flagged as an invalid program in the old form. |
| `7bc8f09` | `new_array(type_tag, dim0, dim1)` pops `dim0 * max(dim1, 1)` values (the array's initial elements) — the decoder was passing `values=()` regardless. Tests updated to push values before `new_array`. |
| `481965e` | `.pyi` docstrings for `await_measure` / `set_detector` / `set_observable` / `return_` now document their stack effects. Also tidies a stale comment in the NewArray rewrite. |
| `e800696` | E2e test upgraded to assert the **full post-rewrite block structure** (7-statement sequence + key attribute values) instead of spot-checking individual elements. Uses `Walk(rule).rewrite(method.code)` — `method` itself isn't an `IRNode` so Walk descends from the top-level `func.Function`. |

All changes verified against the Rust validator's `sim_*` methods in `crates/bloqade-lanes-bytecode-core/src/bytecode/validate.rs` — the single source of truth. 270/270 tests pass.

Other instructions (`initial_fill` / `fill` / `move` / `local_r` / `local_rz` / `global_r` / `global_rz` / `cz` / `get_item` / `dup` / `swap` / `pop` / `halt`) were verified consistent between validator, `.pyi`, and dialect — no fixes needed.

PR #558, opened 2026-04-24
feat(visualize): add entropy search tree visualization

### LINKED ISSUE #557: feat(visualize): add entropy search tree visualization
## Summary

We do not currently have a first-class way to visualize the algorithmic process of entropy-guided search. Debugging the Rust-backed entropy search requires inspecting logs, traces, or internal data structures by hand, which makes it hard to understand why a branch was explored, why a moveset was preferred, or where the search spent its expansion budget.

## Motivation

Entropy search decisions are tree-shaped and iterative: each step expands candidates, scores movesets, updates the active frontier, and eventually converges on a target placement or exhausts a budget. Without a visualization, it is difficult to reason about ranking behavior, compare candidate paths, explain regressions, or tune scoring parameters.

## Proposed Direction

Add a Python visualization surface for entropy search traces that can render the search tree, selected paths, scoring/metric breakdowns, and traversal state in a way that is useful for development and review.

## Acceptance Criteria

- Provide a Python entry point for visualizing entropy search tree traces.
- Surface enough trace state to inspect candidate expansion, scoring, and selected paths.
- Include tests for trace/state/renderer behavior.
- Keep the visualization separate from the core PR that removes the legacy Python search infrastructure.

Jason-Ludmir:
## Summary

- Add an entropy-tree visualization package for inspecting the algorithmic process of Rust-backed entropy search.
- Expose native per-moveset scoring metrics through PyO3 so the visualizer can show score components, closer/further qubits, and selected candidate state.
- Add focused tests for trace conversion, state handling, renderer output, and CLI behavior.

Closes #557.

## Stack

This PR is stacked on top of #556 (`jason/remove-python-search-infra`) so the diff only contains visualization updates and the native scoring hooks they need.

## Test plan

- [x] `uv run pytest python/tests/visualize/entropy_tree -q` → 12 passed
- [x] pre-commit hooks during commit: isort, black, ruff, pyright, cargo fmt, cargo clippy


crates/bloqade-lanes-search/src/entropy.rs
```diff
@@ -709,25 +709,51 @@ pub(crate) fn generate_candidates(
...
+
+    pub fn score(&self, params: &EntropyParams) -> f64 {
+        params.alpha * self.distance_progress
+            + params.beta * (self.arrived as f64)
+            + params.gamma * self.mobility_gain()
+    }
+}
+
+/// Compute the full metrics breakdown for moving from `old_config` to `new_config`.
+///
```
weinbe58, line 742: **[HIGH]** `score_moveset` now allocates two `Vec<u32>` (`closer`, `further`) and sorts them on every call, even though the internal search loop only needs the scalar score and discards the rest. This is a per-call allocation regression on the hot path.

Consider keeping a fast-path `score_moveset` that skips the vecs:
```rust
pub(crate) fn score_moveset(
    old_config: &Config,
    new_config: &Config,
    occupied: &HashSet<u64>,
    ctx: &SearchContext,
    params: &EntropyParams,
) -> f64 {
    // fast path: skip closer/further allocation used only by the visualizer
    // ... (inline the scalar accumulation directly)
}
```
Or make `closer`/`further` opt-in via a flag on `EntropyParams`.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
```diff
@@ -0,0 +1,1121 @@
...
+    """
+    reason = frame.event_reason
+    if reason is None:
+        return None
+    if reason == "state-seen":
+        seen_id = frame.event_state_seen_display_id
+        if seen_id is None:
+            return "entropy bump reason: encountered previously seen configuration"
+        return f"entropy bump reason: seen configuration at display node {seen_id}"
+    if reason == "no-valid-moves":
```
weinbe58, line 45: **[LOW]** Dead branch — `node_w = node_w` is a no-op. The `elif` below already handles the interesting case. Remove this branch (or replace with `pass` and a comment if you want to keep the three-way structure explicit).

python/bloqade/lanes/visualize/entropy_tree/renderer.py
```diff
@@ -0,0 +1,1121 @@
...
+
+
+def _min_horizontal_gap_per_depth(
+    positions: dict[int, tuple[float, float]],
+    nodes: dict[int, NodeState],
+) -> float | None:
+    by_depth: dict[int, list[float]] = defaultdict(list)
+    for node_id, node in nodes.items():
+        if node_id not in positions:
+            continue
```
weinbe58, line 213: **[LOW]** `_ = frame` with the comment "retained for API compatibility with older callers" — but this is new code in a new file, there are no older callers. Remove the `frame` parameter entirely.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
```diff
@@ -0,0 +1,1121 @@
...
+        root_loc = root_configuration.get(qid)
+        target_loc = target.get(qid)
+        if root_loc is not None:
+            positions.append(arch_spec.get_position(root_loc))
+        if target_loc is not None:
+            positions.append(arch_spec.get_position(target_loc))
+
+    if not positions:
+        return full_bounds
+
```
weinbe58, line 118: **[LOW]** `src in source_config.values()` is O(n) in the number of qubits for every lane in the moveset. Build a reverse set once before the loop:
```python
occupied_locs = set(source_config.values())
for lane in moveset:
    src, _dst = arch_spec.get_endpoints(lane)
    if src in occupied_locs:
        ...
```

python/bloqade/lanes/visualize/entropy_tree/state.py
```diff
@@ -0,0 +1,600 @@
...
+import colorsys
+from dataclasses import dataclass, field
+
+from bloqade.lanes.bytecode.encoding import LaneAddress, LocationAddress
+from bloqade.lanes.visualize.entropy_tree.tracer import TreeTraceStep
+
+# Sentinel step_index for synthetic reducer-injected entries (e.g. the
+# root candidate_expand seed that has no corresponding native trace step).
+_SYNTHETIC_STEP_INDEX = -1
+
```
weinbe58, line 14: **[LOW]** `index // 1` is always `index` — this integer floor division by 1 is a no-op. Should be `index % len(sat_cycle)` directly.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
```diff
@@ -0,0 +1,1121 @@
...
+            positions.extend(path)
+            ax.annotate(
+                "",
+                xy=path[-1],
+                xytext=path[-2],
+                arrowprops=dict(arrowstyle="->", color="#C2477F", lw=1.8),
+                zorder=10,
+            )
+    return positions
+
```
weinbe58, line 465: **[HIGH]** `draw_hardware_panel` (~125 lines) is defined here but never imported or called anywhere — not in `controller.py`, `cli.py`, or `__init__.py`. Looks like a leftover from an earlier layout design superseded by `draw_focus_panel`. Should be deleted unless there is a planned use.

python/bloqade/lanes/visualize/entropy_tree/cli.py
```diff
@@ -0,0 +1,186 @@
...
+from bloqade.lanes.visualize.entropy_tree.controller import EntropyTreeController
+from bloqade.lanes.visualize.entropy_tree.state import TreeStateReducer
+from bloqade.lanes.visualize.entropy_tree.tracer import (
+    EntropyTraceBundle,
+    build_entropy_trace,
+    load_kernel_from_file,
+)
+
+
+def _build_default_kernel():  # type: ignore[no-untyped-def]
```
weinbe58, line 17: **[LOW]** `--max-expansions` has no `help=` string. A user unfamiliar with the search algorithm won't know what a "node expansion" is or whether 1000 is a reasonable limit. Suggest:
```python
p.add_argument(
    "--max-expansions",
    type=int,
    default=1000,
    help="Maximum node expansions before stopping the trace. Use -1 for no limit.",
)
```

python/bloqade/lanes/visualize/entropy_tree/state.py
```diff
@@ -0,0 +1,600 @@
...
+    moveset: frozenset[LaneAddress]
+    color: str
+    attempted: bool = False
+    selected: bool = False
+
+
+@dataclass
+class NodeState:
+    node_id: int
+    display_id: int
```
weinbe58, line 41: **[MEDIUM]** `assign_subtree_x` is recursive and Python's default recursion limit (~1000) will be hit for deep search trees (entropy search can produce deep trees with many expansion levels). Either rewrite this iteratively or raise a clear error before the `RecursionError` surfaces.

weinbe58 (commented):
Overall this is a solid piece of work — the tracer/state/renderer/controller separation is clean, the synthetic root-candidate-seed injection in `TreeStateReducer` is a nice touch, and the test coverage for a visualizer is good. A few things worth addressing before merge, ranging from a hot-path performance regression (high priority) to cosmetic dead code.

crates/bloqade-lanes-search/src/entropy.rs
Jason-Ludmir: Handled by restoring `score_moveset` as a scalar fast path. The detailed `closer`/`further` vectors now stay confined to `compute_moveset_metrics`, and a regression test checks score parity without entering the detailed metrics path.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
Jason-Ludmir: Handled by removing the no-op `node_w = node_w` branch while preserving the existing shrink/clamp behavior for dense layouts.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
Jason-Ludmir: Handled by removing the unused `frame` parameter from `_draw_hardware_snapshot` and updating the internal call sites.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
Jason-Ludmir: Handled by building `occupied_locs = set(source_config.values())` once before iterating over the moveset, with a regression test covering the single `values()` call.

python/bloqade/lanes/visualize/entropy_tree/state.py
Jason-Ludmir: Handled by replacing `(index // 1) % len(sat_cycle)` with `index % len(sat_cycle)`.

python/bloqade/lanes/visualize/entropy_tree/renderer.py
Jason-Ludmir: Handled by deleting the unused `draw_hardware_panel`; `draw_focus_panel` is the active hardware-view path.

python/bloqade/lanes/visualize/entropy_tree/cli.py
Jason-Ludmir: Handled by adding help text for `--max-expansions`, including the `-1` no-limit behavior.

python/bloqade/lanes/visualize/entropy_tree/state.py
Jason-Ludmir: Handled by rewriting the tree layout traversal to use an explicit post-order stack instead of recursion, with a 5000-node deep-tree regression test.

weinbe58 (approved):
LGTM

PR #561, opened 2026-04-24
feat(benchmarks): add --arch-spec JSON support to harness

### LINKED ISSUE #548: Add --arch-spec JSON support to benchmark harness
## Problem
The benchmark harness currently relies on the built-in archspec and does not support importing archspec definitions from JSON files.

## Requested changes
- Add a `--arch-spec []` CLI option to import one or more JSON files that define the archspec.
- If `--arch-spec` is not provided, default to the current built-in archspec behavior.
- Update `default_strategy_configs` to accept a closure for archspec selection.
- Wire CLI behavior so lanes can switch between built-in archspec and JSON-provided archspec.

## Notes
This should preserve backward compatibility while enabling external archspec-driven benchmark runs.

Jason-Ludmir:
## Summary
- Adds `--arch-spec PATH [PATH ...]` to the benchmark harness CLI; each JSON file is loaded eagerly and becomes an additional dimension in the case × strategy matrix.
- `default_strategy_configs` now takes an `arch_spec_factory` closure and `arch_spec_id`; the CLI invokes it once per archspec and concatenates strategies.
- `StrategyConfig`/`BenchmarkRow` gain `arch_spec_id`; CSV schema inserts the column after `strategy_id`; diff key is now `(case_id, strategy_id, arch_spec_id)`. Legacy baselines missing the column load as `"builtin"`.

Closes #548

## Test plan
- [x] `uv run pytest python/tests/benchmarks/` — 64 passed
- [x] `uv run black python/benchmarks python/tests/benchmarks --check` — clean
- [x] `uv run isort python/benchmarks python/tests/benchmarks --check-only` — clean
- [x] `uv run ruff check python/benchmarks python/tests/benchmarks` — clean
- [x] `uv run pyright python/benchmarks python/tests/benchmarks` — clean
- [x] CLI smoke-test: `--arch-spec examples/arch/full.json` resolves id `full` end-to-end.
- [ ] Reviewer: spot-check that `--strategies rust_entropy_5 --arch-spec a.json b.json` yields 2× rows per case (matrix expansion).
- [ ] Reviewer: verify legacy baseline compare (pre-column CSV) still loads without error.


weinbe58:
## Review

**Verdict: Approve with minor suggestions.** Well-structured change with thorough tests (64 pass), clean lint/types, and end-to-end smoke test confirms matrix expansion behavior. Closes #548.

### Verified locally
- `uv run pytest python/tests/benchmarks/` → 64 passed
- `uv run black / isort / ruff / pyright` → clean
- CLI smoke test: `--cases ghz_4 --strategies rust_entropy_5 --arch-spec examples/arch/simple.json examples/arch/full.json` → 2 rows tagged `simple` / `full` with `arch_spec_id` column populated correctly. Both reviewer spot-check items in the PR description are empirically verified.

### What works well
- **Eager validation in `_resolve_arch_specs`** (`cli.py:245-284`): JSON parsing and stem-collision detection happen before any benchmark runs — fail fast.
- **Backward-compat for legacy CSVs** (`compare.py:55-62`): old baselines load with `arch_spec_id=\"builtin\"`.
- **Lazy factory closure** (`cli.py:48-55`): `lambda arch=arch: arch` correctly captures per-iteration `arch` via default-arg binding; preserves the legacy lazy-resolution semantics of `default_strategy_configs`.
- **Test coverage** spans matrix expansion, legacy CSV migration, key disambiguation by `arch_spec_id`, per-build factory invocation, stem collisions, malformed JSON, and missing files.
- **CSV column insertion right after `strategy_id`** keeps row identifiers contiguous.

### Suggestions (non-blocking)

1. **`\"builtin\"` is a magic string in four places** — `cli.py:256`, `matrix.py:19`, and the default values on `models.py:34` and `models.py:61`. Pulling into a single module-level `BUILTIN_ARCH_SPEC_ID` constant would prevent drift.

2. **Stem collision with the implicit `\"builtin\"` id is not detected.** A user-provided `--arch-spec path/to/builtin.json` produces `arch_id = \"builtin\"`. The two code paths are mutually exclusive within a single run, but the resulting rows can later be compared against a run from the opposite setup — semantically different archspecs would match by id. Low impact, worth either documenting or reserving the name.

3. **`default_strategy_configs(arch_spec_factory=None, arch_spec_id=\"builtin\")` allows a caller to pass a non-builtin factory while forgetting `arch_spec_id`** — rows would be tagged `\"builtin\"` but built from a different archspec. Consider coupling the two as a single `(id, factory)` argument. The only current caller (the CLI) always passes both, so this is a low-priority footgun.

4. **Behavior change on the default path:** previously, `physical.get_arch_spec()` was effectively called lazily per-strategy build (via `default_strategy_configs`'s factory closure default). Now `_resolve_arch_specs(None)` calls it eagerly once at CLI startup (`cli.py:256`). Fine in practice — just worth noting — but it does mean argparse-error paths now still pay for one `get_arch_spec()` resolution.

5. **Legacy CSV detection at `compare.py:57` is one-version-deep.** Only the immediately-prior schema (no `arch_spec_id`) is recognized. Acceptable while there's only one variant; if a future column lands, this will need to become a list of legacy shapes.

6. **`expand_benchmark_jobs(case_filter=...)` is no longer used by the CLI** (cases are pre-filtered by `select_benchmark_cases`). Tests still exercise it, so it's not dead — but consider trimming it or noting the CLI doesn't drive it anymore.

Jason-Ludmir:
> ### Suggestions (non-blocking)
> 1. **`\"builtin\"` is a magic string in four places** — `cli.py:256`, `matrix.py:19`, and the default values on `models.py:34` and `models.py:61`. Pulling into a single module-level `BUILTIN_ARCH_SPEC_ID` constant would prevent drift.
> 2. **Stem collision with the implicit `\"builtin\"` id is not detected.** A user-provided `--arch-spec path/to/builtin.json` produces `arch_id = \"builtin\"`. The two code paths are mutually exclusive within a single run, but the resulting rows can later be compared against a run from the opposite setup — semantically different archspecs would match by id. Low impact, worth either documenting or reserving the name.
> 3. **`default_strategy_configs(arch_spec_factory=None, arch_spec_id=\"builtin\")` allows a caller to pass a non-builtin factory while forgetting `arch_spec_id`** — rows would be tagged `\"builtin\"` but built from a different archspec. Consider coupling the two as a single `(id, factory)` argument. The only current caller (the CLI) always passes both, so this is a low-priority footgun.
> 4. **Behavior change on the default path:** previously, `physical.get_arch_spec()` was effectively called lazily per-strategy build (via `default_strategy_configs`'s factory closure default). Now `_resolve_arch_specs(None)` calls it eagerly once at CLI startup (`cli.py:256`). Fine in practice — just worth noting — but it does mean argparse-error paths now still pay for one `get_arch_spec()` resolution.
> 5. **Legacy CSV detection at `compare.py:57` is one-version-deep.** Only the immediately-prior schema (no `arch_spec_id`) is recognized. Acceptable while there's only one variant; if a future column lands, this will need to become a list of legacy shapes.
> 6. **`expand_benchmark_jobs(case_filter=...)` is no longer used by the CLI** (cases are pre-filtered by `select_benchmark_cases`). Tests still exercise it, so it's not dead — but consider trimming it or noting the CLI doesn't drive it anymore.

**Implemented in code**

  1. BUILTIN_ARCH_SPEC_ID constant. Added to harness/models.py; the dataclass defaults on StrategyConfig.arch_spec_id and BenchmarkRow.arch_spec_id now reference it, as do compare.py's legacy-CSV migration and
  cli.py's _resolve_arch_specs fallback + --arch-spec help text. No more drifting literals.

  2. Reserve "builtin" stem. _resolve_arch_specs now rejects any --arch-spec path whose filename stem equals BUILTIN_ARCH_SPEC_ID, with a clear error before the duplicate-stem check. This forecloses the cross-run
  id-collision case rather than documenting it. New test: test_resolve_arch_specs_raises_on_reserved_builtin_stem.

  3. Couple id + factory. Replaced default_strategy_configs(arch_spec_factory=..., arch_spec_id=...) with a single arch_spec: tuple[str, Callable[[], ArchSpec]] | None = None. Callers can no longer pass a factory
  while forgetting the id. CLI call site uses arch_spec=(arch_id, lambda arch=arch: arch). Updated test_matrix.py and the three monkeypatched fakes in test_cli.py for the new signature.

  6. Trim case_filter. Removed the unused case_filter parameter from expand_benchmark_jobs. The CLI never forwarded it — cases are pre-filtered upstream by select_benchmark_cases, which also handles size-bucket
  expansion. Docstring updated to explain the resulting asymmetry (only strategy_filter remains because there's no equivalent upstream strategy pre-filter). Test renamed to
  test_matrix_expansion_applies_strategy_filter.

  **Acknowledged, no code change**

  4. Eager vs lazy get_arch_spec() on the default path. The timing change is real (one file-read + JSON-parse moved to CLI startup), but get_arch_spec() is pure today and the alternatives are worse: moving
  resolution back into the per-strategy closure re-introduces N redundant I/O calls on the hot path; conditional lazy-vs-eager branching adds complexity for negligible savings on rare post-resolve parser.error
  paths. Leaving as-is.

  5. One-version-deep legacy CSV detection. Agreed — only the immediate predecessor schema (no arch_spec_id) is recognized. Sufficient while there's exactly one legacy shape. When the next column lands, the
  schema-pinning test will trip and the two ifs become a list of legacy shapes. Not worth pre-engineering.

weinbe58 (approved):

Issue #98, opened 2026-01-09
Generate shot post processing data from `AtomInterpreter` analsys results.


weinbe58:
By convention `ZoneAddress(0)` is the zone that contains all words in the arch spec. As such when the final image is taken there is a bit-string that holds the image results for every single `LocationAddress` in the architecture, whether there is an atom there or not. 

Given a `shot: list[MeasurementResult]` need the following meta-data in order to process the shot results. Note that the `i`th element in `shot` corresponds to the `i`th `LocationAddress` in `ZoneAddress(0)`. 

1.  `logical_mapping: list[list[int]]` such that `logical_mapping[i]` is the list of element indices of `shot` that are the locations of the physical qubits inside the `i`th logical qubit.
2.  `detectors: list[list[int]]` where `detectors[i]` gives the lists of indices of `shot` that should be xor'ed to calculate the `i`th detector. 
3. `observables: list[list[int]]` which is the same as the detectors but just interpreted as the observables of the logical qubits.

weinbe58:
Superseded by #563. The original scope has been narrowed: items 2 & 3 (detectors / observables) are now covered by the static `m2dets` / `m2obs` matrices (`bloqade/lanes/steane_defaults.py` + `logical_mvp.append_measurements_and_annotations`). The detector/observable stripping rewrite lives in [`bloqade-internal#222`](https://github.com/QuEraComputing/bloqade-internal/pull/222) (`RemovePostProcessing`).

What's still genuinely needed — the bridge from raw Zone-0 hardware bitstring to the per-measurement array `generate_post_processing` expects — is tracked in #563 with a tighter, scoped-down work plan. Closing here as duplicate of #563.

PR #565, opened 2026-04-27
feat(atom): get_shot_remapping — full-Zone-0-bitstring → per-logical-qubit indices

### LINKED ISSUE #563: AtomInterpreter shot-remapping: full Zone-0 bitstring → per-measurement array
## Summary

Provide a reusable shot-remapping primitive that converts a raw hardware shot (a full Zone-0 bitstring with one bit per `LocationAddress` in `arch_spec.yield_zone_locations(ZoneAddress(0))`, including empty sites) into a per-`MeasurementResult` array consumable by [`bloqade.gemini.post_processing.generate_post_processing`](https://github.com/QuEraComputing/bloqade-lanes/blob/main/python/bloqade/gemini/post_processing.py).

Replaces the broader scope of #98. After discussion the detector/observable side is now covered by the static `m2dets` / `m2obs` matrices (`bloqade/lanes/steane_defaults.py` + `logical_mvp.append_measurements_and_annotations`), and the post-processing-stripping rewrite lives in [`bloqade-internal#222`](https://github.com/QuEraComputing/bloqade-internal/pull/222) (`RemovePostProcessing`). What's *still* missing — and what neither repo provides today — is the bridge from raw hardware bitstring to the per-measurement `np.ndarray` shape `generate_post_processing` expects.

## Why this matters

`bloqade-internal#222`'s `split_execution_from_postprocessing` produces:

1. an execution kernel (post-`RemovePostProcessing`), and
2. a callable `Callable[[np.ndarray], Iterator[ReturnType]]`.

The callable's input is a 2-D array shaped `(n_shots, n_measurements)`, where the column ordering matches the `TerminalLogicalMeasurement → set_observable / set_detector` IR that `generate_post_processing` introspects via `MeasurementIDAnalysis`. **Hardware delivers shots in a different layout** — a full Zone-0 bitstring per shot, indexed by `LocationAddress` in the architecture's iteration order, *including empty sites*. Without an explicit translation, hooking #222's task layer up to real hardware would silently feed misaligned bits into the post-processing callable.

## Proposed scope

This issue tracks the minimum set of changes needed to make that translation explicit and well-tested:

1. **Extend `MeasureResult` lattice element** in `python/bloqade/lanes/analysis/atom/lattice.py` with a `location_address: LocationAddress` field.
2. **Update `get_future_result_impl`** in `python/bloqade/lanes/analysis/atom/impl.py` to populate the new field from `stmt.location_address` (it already has it locally — just needs to be threaded into the lattice value).
3. ~~**Add `get_shot_remapping(method: ir.Method, arch_spec: ArchSpec) -> list[list[int]]`** as a standalone function in a new module `python/bloqade/lanes/analysis/atom/_shot_remapping.py`. Returns the same `logical_mapping` data structure described in the original #98 — outer index = logical qubit, inner list = indices into the Zone-0 bitstring corresponding to each physical qubit's measurement.~~
   **Update (2026-04-27):** Two design shifts during review:
   - Return shape is **flat** `list[int]` (row-major over the nested `IListResult[IListResult[MeasureResult]]` analysis output) so it indexes directly into the `(n_shots, n_measurements)` post-processing input array. The nested `list[list[int]]` shape was incompatible with the very consumer it exists to serve.
   - Failure handling uses a **typed `ShotMappingResult`** with a structured `ShotRemappingDiagnostic` instead of a bare `None`. A failure here is a compiler-pipeline regression, not a user error — the diagnostic is aimed at compiler developers debugging the failed lowering, with a contextual message and the offending lattice value / location address.
4. **Convenience method on `AtomInterpreter`** that runs the analysis and calls the standalone function.
5. ~~**Export from `python/bloqade/lanes/analysis/atom/__init__.py`**.~~
   **Update (2026-04-27):** Helper module is private (`_shot_remapping`); only `AtomInterpreter.get_shot_remapping` and the result/diagnostic types `ShotMappingResult` / `ShotRemappingDiagnostic` are part of the public surface.
6. **Unit tests** with hand-built lattice values + a small `ArchSpec` (single zone, a few words/sites).
7. **Integration test** via `compile_task()` + a known multi-logical-qubit Steane circuit, asserting the remapping table indexes the right physical qubits.

## Out of scope (separate follow-ups)

- Replacing `m2dets` / `m2obs` static matrices with dynamically-derived equivalents.
- Porting / re-using `RemovePostProcessing` from bloqade-internal#222 once that merges.
- Deprecating the callable-based `PostProcessing` at `analysis/atom/analysis.py:39` (separate cleanup; `generate_post_processing` from `bloqade.gemini.post_processing` is the live path for hardware shots).
- ArchSpec congruence enforcement between hardware and software (latent invariant; orthogonal).

## References

- Supersedes #98 (closing as duplicate of this).
- Consumes / consumed by [`QuEraComputing/bloqade-internal#222`](https://github.com/QuEraComputing/bloqade-internal/pull/222) (Gemini logical device API integration).
- Today's analysis: `MeasureFuture.results: dict[ZoneAddress, dict[LocationAddress, qubit_id]]` already carries enough information to derive the remapping; this issue is mostly about exposing it cleanly.

weinbe58:
## Summary

Closes #563. Adds the bridge between a raw hardware shot (a full Zone-0 bitstring with one bit per `LocationAddress` in `arch_spec.yield_zone_locations(ZoneAddress(0))`, including empty sites) and the per-measurement `np.ndarray` shape `bloqade.gemini.post_processing.generate_post_processing` expects.

This is the durable piece of the original #98 that didn't get superseded by `m2dets` / `m2obs`: hardware shots come back as a full Zone-0 bitstring, and *something* has to project that onto the flat per-measurement array consumed by the post-processing callable produced by `generate_post_processing`. That mapping is what this PR provides.

## Design changes vs. the original spec

Three implementation shifts surfaced during review and are baked into the final PR:

1. **Flat `list[int]` mapping (was `list[list[int]]`).** The original #563 spec called for a list-of-lists indexed by logical qubit, but post-processing takes a flat per-measurement array of shape `(n_shots, n_measurements)`. The flat result indexes directly into the post-processing input array; logical-qubit grouping is recoverable from the kernel itself.
2. **Tagged-union return type (was `list[int] | None` then a single nullable `ShotMappingResult`).** A failure here is a compiler-pipeline regression, not a user error, and the compiler service needs structured debug context — not just `None` and not a string buried in a `RuntimeError`. The result type is now `ShotRemappingOk | ShotRemappingErr`, two independent frozen dataclasses; illegal states are unconstructible.
3. **Default `no_raise=True` on the interpreter wrapper.** `AtomInterpreter.get_shot_remapping(method, *, no_raise=True)` routes through `Forward.run_no_raise`, so an analysis crash falls through into the standard `ShotRemappingErr` path with the `Bottom` lattice as the offending value. Callers see one failure shape; flip `no_raise=False` when debugging an analysis-side bug.

Issue body updated to reflect these shifts.

### Convenience method on `AtomInterpreter`

```python
interp = AtomInterpreter(method.dialects, arch_spec=arch_spec)
match interp.get_shot_remapping(method):
    case ShotRemappingOk(mapping):
        use(mapping)  # list[int] in row-major order
    case ShotRemappingErr(diagnostic):
        log.error("shot remapping failed: %s", diagnostic.message)
```

### Public types

```python
@dataclass(frozen=True)
class ShotRemappingDiagnostic:
    message: str
    offending_value: MoveExecution | LocationAddress

@dataclass(frozen=True)
class ShotRemappingOk:
    mapping: list[int]

@dataclass(frozen=True)
class ShotRemappingErr:
    diagnostic: ShotRemappingDiagnostic
```

All three are re-exported from `bloqade.lanes.analysis.atom` since the compiler service needs to reference them. The standalone `_shot_remapping.get_shot_remapping(return_value, arch_spec)` function stays private — the public surface is the `AtomInterpreter` method; the helper exists so unit tests can drive it with hand-built lattice values without spinning up a kirin method.

## Failure modes (compiler-developer-facing)

`get_shot_remapping` produces a `ShotRemappingErr` carrying a `ShotRemappingDiagnostic` on:

- **Outer / nested-shape mismatch** — `return_value` (or any nested element) is not the expected `IListResult[IListResult[MeasureResult]]` (any other lattice type — `Bottom`, `Unknown`, `Value`, `MeasureFuture`, `MeasureResult`, `DetectorResult`, `ObservableResult`, `TupleResult` — triggers the failure). The diagnostic identifies the failure site by row-major index path (`logical[i]` / `logical[i].physical[j]`).
- **Address outside Zone-0** — a `MeasureResult.location_address` resolves outside `arch_spec`'s Zone-0 iteration, signalling analysis/arch disagreement about hardware layout.

A failure here indicates a pipeline regression rather than a malformed kernel, so the diagnostic is aimed at compiler developers debugging the failed lowering — not surfaced verbatim to users.

## Implementation notes

- **`MeasureResult` lattice element** gains a `location_address: LocationAddress` field. `get_future_result_impl` already had `stmt.location_address` locally — it's now threaded onto the lattice value.
- **`ArchSpec.get_zone_index(loc, zone)`** (O(1) via Rust backend) is used for the Zone-0 index lookup, so we don't materialise a redundant Python dict per call. The helper asserts `len(arch_spec.zones) > 0` at the top so the "address is not in Zone-0" diagnostic stays truthful even if a future arch spec violates the implicit invariant.

## Tests

- **8 unit tests** with hand-built lattice values + a small ArchSpec — cover the happy path (flat output across one or more logical blocks), all four soft-fail conditions (each asserting the diagnostic message identifies the failure site and carries the right offending value), and the empty-list edge cases.
- **1 `@pytest.mark.slow` integration test** compiles a 2-logical-qubit Steane kernel via `compile_squin_to_move(transversal_rewrite=True)` and asserts the resulting flat remapping length matches the analysis output's leaf count (derived from the output, not from the Steane block size), has no overlapping indices, and stays inside the Zone-0 bitstring range.
- Existing `MeasureResult` constructor sites in `test_lattice.py`, `test_atom_interpreter.py`, and `test_gate.py` updated for the new `location_address` arg.

## Test plan

- [x] `uv run pytest python/tests/` — full suite passes.
- [x] Lint / black / isort — clean.

## Out of scope (separate follow-ups)

- Replacing `m2dets` / `m2obs` static matrices with dynamically-derived equivalents.
- Porting / re-using `RemovePostProcessing` from `QuEraComputing/bloqade-internal#222` once that merges.
- Deprecating the callable-based `PostProcessing` at `analysis/atom/analysis.py:39`.
- ArchSpec congruence enforcement between hardware and software.

### DIFF
--- a/python/bloqade/lanes/analysis/atom/__init__.py
+++ b/python/bloqade/lanes/analysis/atom/__init__.py
@@ -1,5 +1,4 @@
 from . import impl as impl
-from ._shot_remapping import get_shot_remapping as get_shot_remapping
 from .analysis import (
     AtomInterpreter as AtomInterpreter,
     PostProcessing as PostProcessing,
--- a/python/bloqade/lanes/analysis/atom/__init__.py
+++ b/python/bloqade/lanes/analysis/atom/__init__.py
@@ -1,4 +1,5 @@
 from . import impl as impl
+from ._shot_remapping import get_shot_remapping as get_shot_remapping
 from .analysis import (
     AtomInterpreter as AtomInterpreter,
     PostProcessing as PostProcessing,
--- a/python/bloqade/lanes/analysis/atom/__init__.py
+++ b/python/bloqade/lanes/analysis/atom/__init__.py
@@ -1,4 +1,8 @@
 from . import impl as impl
+from ._shot_remapping import (
+    ShotMappingResult as ShotMappingResult,
+    ShotRemappingDiagnostic as ShotRemappingDiagnostic,
+)
 from .analysis import (
     AtomInterpreter as AtomInterpreter,
     PostProcessing as PostProcessing,
--- a/python/bloqade/lanes/analysis/atom/__init__.py
+++ b/python/bloqade/lanes/analysis/atom/__init__.py
@@ -1,7 +1,8 @@
 from . import impl as impl
 from ._shot_remapping import (
-    ShotMappingResult as ShotMappingResult,
     ShotRemappingDiagnostic as ShotRemappingDiagnostic,
+    ShotRemappingErr as ShotRemappingErr,
+    ShotRemappingOk as ShotRemappingOk,
 )
 from .analysis import (
     AtomInterpreter as AtomInterpreter,
--- a/python/bloqade/lanes/analysis/atom/_shot_remapping.py
+++ b/python/bloqade/lanes/analysis/atom/_shot_remapping.py
@@ -55,38 +55,34 @@ class ShotRemappingDiagnostic:
 
 
 @dataclass(frozen=True)
-class ShotMappingResult:
-    """Result of computing the Zone-0 bitstring index list for a
-    move kernel's ``terminal_measure`` SSA value.
+class ShotRemappingOk:
+    """Successful shot-remapping result.
 
-    On success ``mapping`` holds the flat list of indices and
-    ``diagnostic`` is ``None``. On failure ``mapping`` is ``None``
-    and ``diagnostic`` carries the debug context.
+    ``mapping`` is the flat list of Zone-0 bitstring indices in
+    row-major order over the input
+    ``IListResult[IListResult[MeasureResult]]`` analysis output.
+    Index directly into the post-processing flat array.
     """
 
-    mapping: list[int] | None
-    diagnostic: ShotRemappingDiagnostic | None = None
+    mapping: list[int]
 
-    @property
-    def ok(self) -> bool:
-        """``True`` iff the mapping was derived successfully."""
-        return self.mapping is not None
 
-    def get(self) -> list[int]:
-        """Return the mapping or raise ``RuntimeError`` (carrying the
-        diagnostic) if the computation failed."""
-        if self.mapping is None:
-            raise RuntimeError(
-                f"ShotMappingResult: mapping unavailable; "
-                f"diagnostic: {self.diagnostic}"
-            )
-        return self.mapping
+@dataclass(frozen=True)
+class ShotRemappingErr:
+    """Failed shot-remapping result.
+
+    ``diagnostic`` carries the contextual message and the offending
+    lattice value or address, aimed at the compiler developer
+    debugging the failed lowering.
+    """
+
+    diagnostic: ShotRemappingDiagnostic
 
 
 def get_shot_remapping(
     return_value: MoveExecution,
     arch_spec: ArchSpec,
-) -> ShotMappingResult:
+) -> ShotRemappingOk | ShotRemappingErr:
     """Project an analysis ``IListResult[IListResult[MeasureResult]]``
     value onto a flat list of Zone-0 bitstring indices.
 
@@ -101,17 +97,23 @@ def get_shot_remapping(
             is flat.
         arch_spec: architecture spec; ``arch_spec.yield_zone_locations(
             ZoneAddress(0))`` defines the canonical Zone-0 bitstring
-            layout that hardware shots are reported against.
+            layout that hardware shots are reported against. Must
+            contain at least one zone (``zone 0`` is the projection
+            target).
 
     Returns:
-        ``ShotMappingResult`` whose ``mapping`` is the flat list of
-        Zone-0 indices in row-major order on success, or ``None`` with
-        a populated ``diagnostic`` on failure. Failure modes:
+        ``ShotRemappingOk`` carrying the flat list of Zone-0 indices
+        on success, or ``ShotRemappingErr`` carrying a
+        ``ShotRemappingDiagnostic`` on failure. Failure modes:
 
-        - ``return_value`` (or any nested element) does not have the
-          expected ``IListResult[IListResult[MeasureResult]]`` shape
-          — most often a sign that the analysis didn't refine the
-          lattice value past ``Bottom`` / ``Unknown``.
+        - ``return_value`` is not an ``IListResult`` (any other
+          ``MoveExecution`` lattice element — ``Bottom``, ``Unknown``,
+          ``Value``, ``MeasureFuture``, ``MeasureResult``,
+          ``DetectorResult``, ``ObservableResult``, ``TupleResult``).
+        - Any element of the outer ``IListResult.data`` is not itself
+          an ``IListResult`` (same set of rejected types as above).
+        - Any element of an inner ``IListResult.data`` is not a
+          ``MeasureResult``.
         - A ``MeasureResult.location_address`` resolves outside
           ``arch_spec``'s Zone-0 iteration — i.e. the analysis and
           arch spec disagree about hardware layout.
@@ -120,11 +122,18 @@ def get_shot_remapping(
         developers, not end users; failures here indicate pipeline
         regressions rather than malformed kernels.
     """
+    # Zone-0 is the projection target. ``ArchSpec.get_zone_index``
+    # returns ``None`` both for addresses outside Zone-0 *and* when
+    # the spec has no Zone-0 at all; assert the second case up front
+    # so the diagnostic's "address is not in Zone-0" wording stays
+    # truthful.
+    assert (
+        len(arch_spec.zones) > 0
+    ), "arch spec invariant violation: no zones (zone 0 expected)"
     zone0 = ZoneAddress(0)
 
     if not isinstance(return_value, IListResult):
-        return ShotMappingResult(
-            mapping=None,
+        return ShotRemappingErr(
             diagnostic=ShotRemappingDiagnostic(
                 message=(
                     "outer return value did not refine to IListResult; "
@@ -137,8 +146,7 @@ def get_shot_remapping(
     remapping: list[int] = []
     for i, logical in enumerate(return_value.data):
         if not isinstance(logical, IListResult):
-            return ShotMappingResult(
-                mapping=None,
+            return ShotRemappingErr(
                 diagnostic=ShotRemappingDiagnostic(
                     message=(
                         f"logical[{i}] did not refine to IListResult; "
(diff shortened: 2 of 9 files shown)


python/bloqade/lanes/analysis/atom/__init__.py
```diff
@@ -1,4 +1,5 @@
 from . import impl as impl
+from ._shot_remapping import get_shot_remapping as get_shot_remapping
```
weinbe58: please remove this import it is not needed here.

weinbe58:
## Adversarial self-review

Tests / lint / pyright clean locally. Comments below are design-level pushback I want to capture before merge.

### 1+2. `ShotMappingResult` allows illegal states, and `.get()` discards structured diagnostic info

`ShotMappingResult.mapping: list[int] | None` and `diagnostic: ShotRemappingDiagnostic | None = None` are independently nullable, with the relationship between them enforced only by docstring convention. The dataclass cheerfully accepts:

- `ShotMappingResult(mapping=None, diagnostic=None)` — `.ok` is `False` and `.get()` raises with `diagnostic: None` baked into the message.
- `ShotMappingResult(mapping=[0, 1], diagnostic=ShotRemappingDiagnostic(...))` — `.ok` is `True`, the diagnostic is silently dropped.

Both states are constructible by anyone reaching for the dataclass directly (and the dataclass is exported in `__init__.py`). A tagged-union shape — two subclasses (`Ok(mapping)` / `Err(diagnostic)`) or a single class with a non-optional discriminator — would enforce the contract at the type level instead of by convention. Minimum-effort fallback: a `__post_init__` validator (`assert (mapping is None) ^ (diagnostic is None)`).

Related: `ShotMappingResult.get()` collapses the structured diagnostic into a `RuntimeError` message string — callers catching the exception can only `str(e)` and parse, not pull `offending_value` back as an object. Either drop `get()` and force callers through `.ok` / `.diagnostic`, or raise a custom exception that carries the diagnostic as an attribute. If we go with the tagged-union refactor, `Ok.unwrap()` is safe by construction and this part dissolves.

### 3. Naming inconsistency — `ShotMappingResult` vs `ShotRemappingDiagnostic` vs `get_shot_remapping`

Function is `get_shot_remapping`, diagnostic is `ShotRemappingDiagnostic`, but the result is `ShotMappingResult` (no "re-"). Three places, two spellings. Both types are exported from `bloqade.lanes.analysis.atom`, so the asymmetry shows up in every downstream import. Cheap to fix now, irritating later. Suggest `ShotRemappingResult` for consistency.

### 6. `get_shot_remapping` docstring narrows the failure mode

The docstring says:

> ``return_value`` (or any nested element) does not have the expected shape — most often a sign that the analysis didn't refine the lattice value past ``Bottom`` / ``Unknown``.

The diagnostic itself reports any non-`IListResult` (`Value`, `MeasureResult`, `TupleResult`, `MeasureFuture`, …). A reader could plausibly walk away thinking `Bottom` / `Unknown` are the only triggers. Tighten to enumerate all rejected cases (or at least say "any non-`IListResult`").

Meta point: treat this docstring as the contract surface. Whenever the rejection rules change — not just when the typical-cause hint goes stale — update the docstring along with the code. The "most often" hedge will silently rot otherwise.

### 8. Integration test hard-codes `num_logical * 7`

```python
assert len(remapping) == num_logical * 7
```

The `7` is the Steane [[7,1,3]] block size baked in as a magic number. If `compile_squin_to_move(transversal_rewrite=True)` ever switches code or default, the test fails for the wrong reason — reads as "Steane block size is wrong" instead of "shot remapping disagreed with the kernel's measurement count." Derive the expected length from the analysis output instead:

```python
_, output = interp.run(physical_move)
expected_len = sum(len(logical.data) for logical in output.data)
assert len(remapping) == expected_len
```

The contract under test is "the remapping length equals the number of `MeasureResult` leaves" — that's exactly what the function promises, and it doesn't tie the assertion to Steane parameters.

### 9. Soft-fail vs. interpreter-exception boundary is implicit; use `run_no_raise` + `no_raise` flag

`AtomInterpreter.get_shot_remapping(method)` has two failure shapes today:

1. `self.run(method)` raises (analysis bug) — caller sees a Python exception, no `ShotMappingResult` involved.
2. The analysis completes but the lattice value didn't refine, or a `MeasureResult.location_address` resolves outside Zone-0 — caller gets `ShotMappingResult` with `ok=False`.

Both happen for "compiler-pipeline regression" reasons but require different handling at the call site. Swap `self.run(method)` for `self.run_no_raise(method)` and add a `no_raise: bool = True` flag:

```python
def get_shot_remapping(
    self, method: ir.Method, *, no_raise: bool = True
) -> _shot_remapping.ShotMappingResult:
    run_method = self.run_no_raise if no_raise else self.run
    _, output = run_method(method)
    return _shot_remapping.get_shot_remapping(output, self.arch_spec)
```

`run_no_raise` is the kirin-provided helper that wraps `run` in `try/except Exception` and substitutes `(initial_frame, lattice.bottom())`, so an analysis crash falls through into the existing "outer return value did not refine to IListResult; got Bottom" diagnostic. The contract collapses to a single shape: `ShotMappingResult` either holds a mapping or a diagnostic. Callers debugging an analysis crash that needs the traceback flip to `no_raise=False`.

Matches the convention already used elsewhere in the analysis stack: `upstream.py:117,129`, `transform.py:59`, `gemini/logical/group.py:71`.

Tradeoff: when `no_raise=True`, an analysis crash and a legitimate "Bottom output from a clean run" produce identical diagnostics. The `no_raise=False` escape hatch covers the debug case.

### 10. Add an assert for the Zone-0 invariant

`get_shot_remapping` calls `arch_spec.get_zone_index(loc, ZoneAddress(0))` and treats `None` as "address not in Zone-0." But that branch *also* fires when the arch spec has no Zone-0 at all (every call returns `None`), producing a misleading "address is not in Zone-0" diagnostic.

The spec invariant ("if there's any zone, there's a zone 0") makes this unreachable in practice, but an explicit assert at the top makes the invariant load-bearing instead of implicit:

```python
zone0 = ZoneAddress(0)
assert any(True for _ in arch_spec.yield_zone_locations(zone0)), \
    "arch spec invariant violation: no Zone-0"
```

Cheap, explicit, fails with a useful message if an arch spec ever violates the invariant.

---

Out of scope for this PR but worth tracking: `MeasureResult.copy()` is shallow over `location_address` while `IListResult.copy()` deep-copies — confirmed intentional (LocationAddress is value-semantic), no action needed here.

Issue #443, opened 2026-04-09
refactor(python): merge bloqade.lanes.arch and bloqade.lanes.layout packages


weinbe58:
## Problem

`bloqade.lanes.arch` and `bloqade.lanes.layout` are tightly coupled but split across two packages:

- `layout/` has: `ArchSpec`, `LocationAddress`, `LaneAddress`, `PathFinder`, `Word`, `MoveMetricCalculator`, encoding
- `arch/` has: `ZoneBuilder`, `ArchBuilder`, `build_arch`, topologies, `ZoneSpec`, `WordGrid`

Users constantly import from both. Internal code crosses the boundary frequently (e.g. builder imports from layout, layout.arch imports from bytecode._native).

## Proposal

Brainstorm a merged package structure that:
- Groups related concepts together (addresses, arch spec, builders, path finding)
- Reduces import confusion
- Keeps backward-compatible re-exports during migration

Candidate structure to explore:
```
bloqade.lanes.arch/
  __init__.py        # public API re-exports
  spec.py            # ArchSpec wrapper
  encoding.py        # LocationAddress, LaneAddress, Direction, MoveType
  builder.py         # ZoneBuilder, ArchBuilder, build_arch
  topology.py        # SiteTopology, WordTopology, InterZoneTopology impls
  zone.py            # ZoneSpec, ArchBlueprint, DeviceLayout
  word.py            # Word, WordGrid
  path.py            # PathFinder, MoveMetricCalculator
  gemini/            # device-specific specs
```

## Notes

- Follow-up to the zone-centric archspec branch
- Should not change Rust-side code
- Old import paths can be preserved via re-exports during transition

weinbe58:
Superseded by #569, which has the revised structure (foundation types in  rather than under ) and a concrete 9-stage migration plan.

weinbe58:
(Re-posting; the previous close-comment had backticks eaten by the shell.)

Superseded by #569, which has the revised structure (foundation types in `bloqade.lanes.bytecode/` rather than under `bloqade.lanes.arch/` as originally proposed here) and a concrete 9-stage migration plan with naming-consistency renames folded in per stage.

Issue #419, opened 2026-04-03
feat(arch): add zone-level ArchSpec APIs for flair AOD move integration


weinbe58:
## Summary

Add zone-addressed APIs to `ArchSpec` that expose move and grid information in a form natural to flair's AOD waypoint abstraction. This enables flair to generate AOD grid moves directly from the ArchSpec, making it the source of truth for the flair standard library.

**Builds on:** #398 (zone-based architecture builder)
**Design doc:** `plans/2026-04-03-archspec-flair-bridge-design.md`

## New APIs

Four new methods on `ArchSpec`:

1. **`get_zone_grid(zone_id) -> Grid`** — all site positions in a zone as a `bloqade.geometry.Grid`
2. **`get_all_sites() -> list[tuple[float, float]]`** — canonical flat site list across all zones (column-major per `bloqade-geometry` convention)
3. **`get_available_buses(zone_id) -> list[BusDescriptor]`** — enumerate valid `(bus_id, move_type, direction)` for a zone
4. **`get_grid_endpoints(zone_id, bus_id, move_type, direction) -> tuple[Grid, Grid]`** — start/end grids for a bus move at full occupancy

## Internal restructuring

- Zone-indexed bus lookups (`_zone_site_buses`, `_zone_word_buses`) built during `__post_init__`

## Migration plan

1. **Step 1 (#420):** Add zone-addressed APIs, restructure internals. Current compiler continues using Zone 0.
2. **Step 2 (#421):** Refactor Kirin compiler to use zone-addressed APIs. Add zone index to `LaneAddress`/`LocationAddress`.

## Work plan

- [x] Restructure ArchSpec internals to zone-indexed bus data models (builds on #398) — done in #437 (zone-scoped bus data in Rust `Zone` struct, accessed via `self._inner.zones[i].site_buses/word_buses`)
- [ ] Implement `get_zone_grid(zone_id)`
- [ ] Implement `get_all_sites()`
- [ ] Implement `get_available_buses(zone_id)`
- [ ] Implement `get_grid_endpoints(zone_id, bus_id, move_type, direction)`
- [x] Unit tests with small hand-built ArchSpec — done in #437 (51 tests in `test_arch_builder.py`)
- [x] Integration test with Gemini arch via `build_arch()` — done in #437 (7 tests in `test_spec.py`)
[...]

PR #567, opened 2026-04-27
refactor(arch): trim transitional Python-only API surface from ArchSpec

weinbe58:
## Summary

Trim transitional Python-only API surface from `bloqade.lanes.layout.ArchSpec` so the wrapper stays close to a thin shell over the underlying Rust spec, in preparation for eventually removing the wrapper class entirely (refs #443).

Four self-contained cleanups, one commit each.

### 1. Extract zone-addressed APIs into `ArchSpecGeometry` (`9bade05`)

The `#419/#420`-era helpers (`get_zone_grid`, `get_all_sites`, `get_available_buses`, `get_grid_endpoints`) plus `BusDescriptor` are Python-side computation that produces `bloqade.geometry` grids and Python-friendly bus descriptors — they don't belong on the core `ArchSpec` wrapper. Moved into a new `ArchSpecGeometry` helper class in `python/bloqade/lanes/layout/arch_geometry.py`.

`ArchSpecGeometry` accepts either a Python `ArchSpec` or the underlying Rust `_RustArchSpec`, so downstream consumers can build it from either layer:

```python
from bloqade.lanes.layout.arch_geometry import ArchSpecGeometry
geom = ArchSpecGeometry(arch_spec)  # or ArchSpecGeometry(rust_spec)
geom.get_zone_grid(0)
```

### 2. Drop visualization shims (`06f0a5f`)

Five forwarding shims (`plot`, `show`, `path_bounds`, `x_bounds`, `y_bounds`) and the `_visualizer` cached_property on `ArchSpec` delegated to `bloqade.lanes.visualize.arch.ArchVisualizer`. Dropped — downstream builds the visualizer explicitly:

```python
from bloqade.lanes.visualize.arch import ArchVisualizer
ArchVisualizer(arch).plot(...)
```

Migrated `visualize/artist.py`, two demo files, and three test files.

### 3. `get_position` / `get_endpoints` → `Optional` (`68f7f29`)

Both methods previously raised `ValueError` on a `None` Rust result; their Rust counterparts (`location_position`, `lane_endpoints`) return `Optional`. The drift would silently swallow None values when the Python wrapper is eventually replaced with the Rust spec.

Changed both methods to return `Optional`. Migrated ~30 callers via inline `assert ... is not None` (where the caller knows the address is valid) or upward `None` propagation.

### 4. Rewrite `yield_zone_locations` docstring (`e75052a`)

Old docstring framed the iterator as yielding locations "in a given zone address" (suggesting it filters words by zone) and carried a stale "Python heuristic layer" historical paragraph. The actual contract is the canonical iteration that `get_zone_index(loc, zone_address)` numbers — every `(word, site)` pair tagged with the requested zone, in word-major / site-major order. Updated the docstring to say this directly.

## Breaking changes

**Python API only** (Rust + C unchanged):

- `ArchSpec.get_zone_grid` / `get_all_sites` / `get_available_buses` / `get_grid_endpoints` removed — use `ArchSpecGeometry(arch).<method>(...)`.
- `BusDescriptor` no longer re-exported from `bloqade.lanes.layout`; import from `bloqade.lanes.layout.arch_geometry` instead.
- `ArchSpec.plot` / `show` / `path_bounds` / `x_bounds` / `y_bounds` removed — use `ArchVisualizer(arch).<method>()`.
- `ArchSpec.get_position` / `get_endpoints` now return `Optional`; callers must handle `None` (existing behavior was to raise `ValueError`).

Pre-release; no backport.

## Test plan

- [x] `uv run pytest python/tests/` — full suite green: **1097 passed, 9 skipped**.
- [x] Lint / black / isort — clean.
- [x] Demos smoke-tested with `MPLBACKEND=Agg`.

## Out of scope (deferred to follow-ups)

- Drop `site_buses` / `word_buses` aggregates after `search/scoring.py` removal lands.
- `_inner` access patterns elsewhere (`MoveSolver.from_arch_spec`, atom-state Rust glue) — natural migration when `RustWrapper` is removed.
- Parameter renames (`loc_addr` → `location_address`, `zone_id: ZoneAddress` → `zone_address`), `word_zone_map` docstring sharpening, "home" terminology disambiguation — folded into the package-merge PR (refs #443).

### DIFF
--- a/python/bloqade/lanes/layout/arch_geometry.py
+++ b/python/bloqade/lanes/layout/arch_geometry.py
@@ -1,0 +1,226 @@
+"""Geometry-level query helpers for ArchSpec.
+
+This module provides :class:`ArchSpecGeometry`, a helper class for downstream
+consumers who want geometry-level queries (grids, flat site lists, bus
+descriptors) without walking Rust zone objects directly.
+
+The class accepts either a Python :class:`~bloqade.lanes.layout.arch.ArchSpec`
+wrapper or a raw Rust ``_RustArchSpec`` object, so it can be used at either
+layer of the stack.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import TYPE_CHECKING
+
+from bloqade.lanes.bytecode._native import ArchSpec as _RustArchSpec
+from bloqade.lanes.layout.encoding import Direction, LaneAddress, MoveType
+
+if TYPE_CHECKING:
+    from bloqade.geometry.dialects.grid import Grid as GeoGrid
+
+    from bloqade.lanes.layout.arch import ArchSpec
+
+
+@dataclass(frozen=True)
+class BusDescriptor:
+    """Descriptor for a bus within a zone."""
+
+    bus_id: int
+    move_type: MoveType
+    direction: Direction
+    num_lanes: int
+
+
+class ArchSpecGeometry:
+    """Geometry-level query helper wrapping an ArchSpec.
+
+    Provides methods for retrieving coordinate grids, flat site lists, and bus
+    descriptors from an architecture specification. Intended for downstream
+    consumers who need geometry-level data without walking Rust zone objects
+    directly.
+
+    Accepts either a Python :class:`~bloqade.lanes.layout.arch.ArchSpec` or a
+    raw Rust ``_RustArchSpec`` object, so it can be constructed from either
+    layer of the stack.
+    """
+
+    def __init__(self, arch_spec: "ArchSpec | _RustArchSpec") -> None:
+        if isinstance(arch_spec, _RustArchSpec):
+            self._inner = arch_spec
+        else:
+            self._inner = arch_spec._inner
+
+    def get_zone_grid(self, zone_id: int) -> GeoGrid:
+        """Get the coordinate grid for a zone as a ``bloqade.geometry.Grid``.
+
+        Args:
+            zone_id: Zone index.
+
+        Returns:
+            A ``bloqade.geometry.dialects.grid.Grid`` with the zone's
+            x and y positions.
+
+        Raises:
+            ValueError: If zone_id is out of range.
+        """
+        from bloqade.geometry.dialects.grid import Grid as GeoGrid
+
+        if zone_id < 0 or zone_id >= len(self._inner.zones):
+            raise ValueError(
+                f"zone_id {zone_id} out of range [0, {len(self._inner.zones)})"
+            )
+        zone = self._inner.zones[zone_id]
+        return GeoGrid.from_positions(
+            tuple(zone.grid.x_positions), tuple(zone.grid.y_positions)
+        )
+
+    def get_all_sites(self) -> list[tuple[float, float]]:
+        """Get all site positions across all zones in canonical order.
+
+        Returns positions in zone-major order, with each zone flattened
+        in column-major grid order (``x`` outer, ``y`` inner).
+        Each position is an ``(x, y)`` tuple.
+        """
+        sites: list[tuple[float, float]] = []
+        for zone in self._inner.zones:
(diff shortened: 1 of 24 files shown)


python/bloqade/lanes/layout/arch_geometry.py
```diff
@@ -0,0 +1,226 @@
...
+            direction: FORWARD or BACKWARD.
+
+        Returns:
+            ``(src_grid, dst_grid)`` where each grid contains the physical
+            positions of all source/destination sites for this bus.
+
+        Raises:
+            ValueError: If zone_id or bus_id is out of range, or
+                move_type is not SITE or WORD.
+        """
```
weinbe58, line 165: Verify that this is not already a method in the Rust implementation of the ArchSpec.

python/bloqade/lanes/search/tree.py
```diff
@@ -109,7 +109,11 @@ def _build_lane_indexes(self) -> None:
                                         direction,
                                         zone_id,
                                     )
-                                    src, _ = self.arch_spec.get_endpoints(lane)
+                                    _ep = self.arch_spec.get_endpoints(lane)
+                                    assert (
+                                        _ep is not None
+                                    ), f"lane {lane!r} has no endpoints in this architecture"
+                                    src, _ = _ep
```
weinbe58: go back and keep the old behavior of `arch_spec.get_endpoints` and implement a non-throwing variant called `arch_spec.try_get_endpoints` which matches the current rust implementation. The man reason is to keep these changes to a minimum in search as this module is going to be deleted in pr #556

weinbe58:
Addressed in `137fe8a` and `6fb8105`.

**@weinbe58 #1 — Rust equivalent for `get_grid_endpoints`:** verified `_native.pyi` — no `grid_endpoints` / `bus_grid` / equivalent method on the Rust `ArchSpec`. Only `lane_endpoints` exists. So `ArchSpecGeometry.get_grid_endpoints` is genuinely Python-side aggregation (walks the bus's lanes via `lane_endpoints` and groups the resulting positions into source/destination grids). Nothing to delegate to today.

**@weinbe58 #2 — restore raising behaviour on `get_*` and add `try_get_*` siblings:** done in `137fe8a`.
- `ArchSpec.get_position` / `ArchSpec.get_endpoints` restored to raising `ValueError` on `None` (matches the original behaviour on `main`).
- New `ArchSpec.try_get_position` and `ArchSpec.try_get_endpoints` siblings return `Optional`, matching the Rust `location_position` / `lane_endpoints` contract.
- Reverted every caller migration introduced by `68f7f29` (search/* + non-search). Diffs in `search/tree.py`, `scoring.py`, generators, traversal, and the rewrite/heuristic/metrics/artist callers all go back to direct unpacks.
- `test_arch_extra.py` now covers four cases: `try_get_*` returning `None` and `get_*` raising `ValueError` for invalid addresses.

**Copilot — actionable findings:** addressed in `6fb8105`.
- `arch_geometry.get_grid_endpoints`: `zip(bus.src, bus.dst, strict=True)` matches the codebase's existing strict-zip convention; `lane_endpoints` / `location_position` returning `None` now raise `ValueError` instead of silently producing partial grids.
- `visualize/artist.StateArtist`: now caches one `ArchVisualizer` per instance via a `cached_property`, replacing the three inline constructions in `_show_global` / `show_cz` / `show_slm` (preserves per-instance bounds caching).

**Copilot — non-issues after the pivot:**
- `arch.py:343` `assert` complaint: moot — `get_path` reverted to direct unpacks since `get_endpoints` raises again.
- `artist.py:540` `draw_atoms` Optional handling: moot — `get_position` raises again.
- `demo/pipeline_details.py:101` (suppressed) demo indexing: moot for the same reason.

Full suite green: 1099 passed, 9 skipped; pyright + ruff clean.

PR #590, opened 2026-04-29
fix(benchmarks): avoid Initialize validation failure

### LINKED ISSUE #589: fix(benchmarks): avoid Initialize validation failure in physical benchmarks
## Summary
Physical benchmark rows that previously compiled now fail after the recent main merge with `_ValidationAnalysis does not support Initialize() using logical dialect`.

## Root Cause
Eager `NewAt` address validation runs after `_RewriteU3ToInitialize()` can introduce Gemini logical `Initialize` statements. The validation interpreter is intended for Lanes move-address validation plus `gemini.common.NewAt`, so it cannot interpret Gemini logical `Initialize`.

## Expected Fix
Validate explicit `NewAt` addresses before introducing logical `Initialize`, and add a regression test covering a kernel with `squin.u3` so the unsupported-dialect error does not return.

Jason-Ludmir:
## Summary
- Run eager explicit-allocation validation before logical initialization rewrites introduce Gemini `Initialize` statements.
- Add a regression test covering the physical noise/fidelity compile path for `squin.u3` kernels.

## Test plan
- `uv run --python /Users/jasonludmir/.local/bin/python3.12 --locked pytest python/tests/gemini/validation/test_e3_pipeline_validation.py -v`
- `uv run --python /Users/jasonludmir/.local/bin/python3.12 --locked pytest python/tests/test_compile_api_split.py python/tests/gemini/validation/test_e3_pipeline_validation.py -v`
- `uv run --python /Users/jasonludmir/.local/bin/python3.12 --locked python -m benchmarks.cli --architecture physical --cases adder_4,qpe_9,trotter_rand_35 --strategies rust_astar --output /tmp/benchmark-init-fix.csv`

Fixes #589

### DIFF
--- a/python/bloqade/lanes/upstream.py
+++ b/python/bloqade/lanes/upstream.py
@@ -42,6 +42,21 @@ class NativeToPlace:
 
     def emit(self, mt: Method, no_raise: bool = True):
         out = mt.similar(mt.dialects.add(place))
+        if self.arch_spec is not None:
+            from bloqade.gemini.analysis.duplicate_address_validation import (
+                DuplicateAddressValidation,
+            )
+            from bloqade.lanes.validation.address import Validation
+
+            errors: list[ir.ValidationError] = []
+            _, per_stmt_errors = Validation(arch_spec=self.arch_spec).run(out)
+            errors.extend(per_stmt_errors)
+            _, dup_errors = DuplicateAddressValidation().run(out)
+            errors.extend(dup_errors)
+            if errors:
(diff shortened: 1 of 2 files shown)


weinbe58:
I am going to be doing a big refactor of the compiler pipeline so this issue will be resolved later today if that works for you @Jason-Ludmir?

Jason-Ludmir:
> I am going to be doing a big refactor of the compiler pipeline so this issue will be resolved later today if that works for you @Jason-Ludmir?

That sounds good @weinbe58

PR #595, opened 2026-05-01
feat(heuristics): no-return placement family + loose-goal Rust solver

### LINKED ISSUE #564: Expose Rust-bound LaneAddress / MoveStep from MoveSolver.solve (replace positional 6-tuple move_layers)
### Context

Follow-up from #556 (refactor that removed Python search infrastructure and made the Rust `MoveSolver` the sole search backend) and #527.

### Problem

`MoveSolver.solve` already returns `goal_config` as `dict[int, _native.LocationAddress]` — a PyO3-wrapped struct accessed via `.word_id` / `.site_id` / `.zone_id`. Good.

But `result.move_layers` is still:

```
list[list[tuple[int, int, int, int, int, int]]]
```

— a list of layers, each a list of raw 6-tuples decoded **positionally** as `(direction, move_type, zone, word, site, bus)`. The decoder lives in `python/bloqade/lanes/heuristics/physical/movement.py`:

```python
_DIR_MAP = {0: Direction.FORWARD, 1: Direction.BACKWARD}
_MT_MAP = {0: MoveType.SITE, 1: MoveType.WORD, 2: MoveType.ZONE}

def convert_move_layers(...):
    ...
    for d, mt, zone, word, site, bus in step:
        LaneAddress(_MT_MAP[mt], word, site, bus, _DIR_MAP[d], zone)
```

This positional decoding is fragile:

- Any reshuffle of fields on the Rust side breaks decoding silently — the unpack still succeeds but assigns wrong fields to the `LaneAddress`.
- `_DIR_MAP` / `_MT_MAP` are int-tag translation tables that duplicate enum knowledge already exposed by the existing `bloqade.lanes.bytecode._native.{Direction,MoveType}` PyO3 enums.
- This class of bug has bitten us before: #510 was a missed `MoveType.ZONE = 2` entry in `_MT_MAP`. The fix is currently a regression test (`test_cz_placements_rust_handles_zone_move_type`), but the underlying brittleness remains as long as the int-tag translation layer exists.

### Proposal

Mirror the existing `LocationAddress` PyO3 wrapping pattern:

- Add a Rust-side `LaneAddress` (or `MoveStep`) struct exposed via PyO3, with named attributes (`move_type`, `word_id`, `site_id`, `bus_id`, `direction`, `zone_id`).
- Have `MoveSolver.solve` return `move_layers` as `list[list[_native.LaneAddress]]`.
- Delete `_DIR_MAP` / `_MT_MAP` and the positional unpack in `convert_move_layers`. The function shrinks to a one-liner that maps `_native.LaneAddress` → `bloqade.lanes.layout.LaneAddress` (or, if the Rust type is sufficient, callers consume it directly and `convert_move_layers` disappears entirely).

### Acceptance

- `move_layers` callers in `python/bloqade/lanes/heuristics/physical/movement.py` and `python/bloqade/lanes/heuristics/move_synthesis.py` no longer reference `_DIR_MAP` / `_MT_MAP`.
- `convert_move_layers` either disappears or becomes trivial.
- `test_cz_placements_rust_handles_zone_move_type` (regression for #510) still passes — variant coverage is now structural rather than dependent on a translation table.

### References

- #556 — removed Python search infrastructure (current state of the decoder)
- #527 — closing issue for #556

lsschmid:
## Summary

Introduces a **no-return placement strategy family** — three new physical placement strategies for the Gemini logical pipeline that chain CZ layers together without returning atoms to home positions between gates.

In main today, `PhysicalPlacementStrategy` requires Python to pre-compute target locations for every qubit per CZ layer (via the `TargetGeneratorABC` plugin). The Rust solver then routes atoms to those targets. With "no-return" placement, Python only specifies **which qubits should pair up for CZ gates**; the Rust solver picks the actual entangling positions itself via Hungarian-driven optimization. Skipping the return-to-home moves saves wall-clock cost at the price of an extra Rust-side optimization step per layer — a net win at high atom density.

## The three strategies

| Strategy | What it does | When to reach for it |
|---|---|---|
| `NoReturnPlacementStrategy` | Runs a single Hungarian per CZ layer (optionally lookahead-aware across future CZ layers) to pick entangling positions, then routes current → picked positions in one Rust call. | The baseline. Fast and competitive at most densities. |
| `NoHomePlacementStrategy` | Two-phase Hungarian: first picks intelligent "home" positions (with gamma-decayed future-CZ proximity), then routes home → entangling sites in a fixed-target second phase. | Workloads where caching home positions between layers helps — e.g. circuits where the same qubits gate repeatedly with the same partners. |
| `RecedingHorizonNoReturnPlacementStrategy` | MPC-style: generates K diverse Hungarian candidates per layer, simulates each forward briefly, commits the best one's path, re-plans. | High atom density (≥80q, depth ≥3) where the single-Hungarian assignment under-uses parallelism. |

All three are opt-in. `PhysicalPlacementStrategy` is the default and remains untouched.

## Python integration

All four placement strategies share the same Python contract (`PlacementStrategyABC`):

```
PlacementStrategyABC                         (existing, in analysis/placement/)
├── PhysicalPlacementStrategy                # existing — Python picks targets via TargetGeneratorABC plugin
└── NoReturnStrategyBase                     # new abstract base for the no-return family
    ├── NoReturnPlacementStrategy            # → solver.solve_entangling()
    ├── NoHomePlacementStrategy              # → solver.solve_nohome()
    └── RecedingHorizonNoReturnPlacementStrategy  # → solver.solve_entangling_rh()
```

`NoReturnStrategyBase` owns the shared scaffolding (solver caching, `SolveOptions` construction, the `cz_placements` request/response plumbing, the standard `sq_placements` / `measure_placements` paths, lattice conversion, and the `rust_nodes_expanded_total` observability counter). Each subclass implements **one method** — `_invoke_solver(...)` — calling its specific Rust solver entry point.

The `strategy` and `deadlock_policy` fields are typed as `_native.SearchStrategy` / `_native.DeadlockPolicy` enums (not strings), so typos fail at construction rather than as a runtime `KeyError`.

**Python-side classes used**: `PlacementStrategyABC` (contract), `NoReturnStrategyBase` (scaffolding), `AtomState` / `ConcreteState` / `ExecuteCZ` / `ExecuteMeasure` (existing lattice types), `MoveSolver` (PyO3 binding, cached per strategy instance via `MoveSolver.from_arch_spec(...)`), `_native.SolveOptions` + strategy-specific `EntanglingOptions` / `NoHomeOptions` / `RecedingHorizonOptions`.

**Handover to Rust** happens inside `NoReturnStrategyBase.cz_placements`. Python encodes:

- `initial: dict[qubit_id → LocationAddress._inner]`
- `cz_pairs: list[(control_qid, target_qid)]`
- `blocked: list[LocationAddress._inner]` (spectator atom positions)
- `future_cz_layers: list[list[(c, t)]] | None` (for lookahead-aware Hungarian)

The subclass's `_invoke_solver(...)` calls the appropriate `MoveSolver.solve_*()` method through PyO3 and returns a `SolveResult`. The base then converts `SolveResult.goal_config` + `SolveResult.move_layers` back into an `ExecuteCZ` lattice element for downstream rewrite passes.

`PalindromePlacementStrategy(inner=...)` works as a wrapper around any of these — the contract is preserved.

## Backward compatibility

**`HeuristicGenerator` (shared by all search strategies)** gained three behaviors during this work. Two are gated on `ctx.cz_pairs.is_some()` so plain `solve()` / `astar` / `ids` / `cascade-ids` callers see **byte-identical heuristic-generator output**:

- Contested-destination soft penalty (`generators/heuristic.rs:348`)
- Fallback `truncate(3)` vs. `truncate(1)` when no candidate has a positive score (`generators/heuristic.rs:438`)

The third (accidental-CZ spectator-escape detection) is implicitly gated — its loop body never executes for fixed-target solves where every qubit is in `ctx.targets`.

`Strategy::Entropy` callers — including the default `PhysicalPlacementStrategy(traversal=RustPlacementTraversal(strategy="entropy"))` — don't touch `HeuristicGenerator` at all, so they're unaffected regardless.

**`IdsFrontier` ordering** *is* changed for all IDS callers (not gated): the frontier is now h-primary with depth as tiebreaker, plus a reversal-aware penalty. These are bug fixes for h-plateau cycles that previously caused infinite loops on certain topologies (commits `6821b99`, `f941ded`). The IDS trajectory shifts slightly for every IDS consumer; the only observable impact in the benchmark suite was a 4-node `nodes_explored` bump on `trotter_rand_35/rust_ids` (620 → 624), no quality regression — baseline updated in this PR.

## Files / structure

**Rust** (`crates/bloqade-lanes-search/src/`):

- `entangling.rs` (new) — Hungarian variants: vanilla, congestion-aware, occupancy-penalty, multi-layer lookahead, iterative w/ Case A/B/D/E displacement. Also owns the shared cost-matrix constants (`MOVE_PENALTY`, `LOOKAHEAD_BETA`, `OCCUPANCY_PENALTY_DEFAULT`).
- `nohome.rs` (new) — two-phase Hungarian with bus-signature reward
- `receding_horizon.rs` (new) — K-candidate rollout orchestrator with greedy-first beam + IDS fallback
- `goals.rs` — adds `EntanglingConstraintGoal` and `PartialPlacementGoal`
- `generators/loose_target.rs` (new) — per-restart seeded target generator
- `solve.rs` — `SolveOptions` split into `SolveOptions` / `EntropyOptions` / `EntanglingOptions` (defaults preserved across the split)
- `heuristic.rs` — `DistanceTable` now uses a 2D `Vec<u32>` indexed via a side `HashMap<u64, usize>` for better cache locality on the Hungarian hot path (was `HashMap<(u64,u64), u32>`)

**Python** (`python/bloqade/lanes/heuristics/physical/`):

- `_solver_dispatch.py` (new) — shared `_STRATEGY_MAP` (legacy string-coercion for `PhysicalPlacementStrategy.traversal`)
- `_no_return_base.py` (new) — `NoReturnStrategyBase` ABC (enum-typed `strategy`/`deadlock_policy` fields)
- `no_return.py` (renamed from `loose_goal.py`) — `NoReturnPlacementStrategy`
- `nohome.py` — `NoHomePlacementStrategy`
- `receding_horizon.py` — `RecedingHorizonNoReturnPlacementStrategy`

**Design doc**: `docs/superpowers/plans/2026-05-11-receding-horizon-loose-goal-design.md`.

## Test plan

- [x] `cargo test --workspace` — 245 in search crate + 228 in bytecode-core all pass
- [x] `uv run maturin develop --release && uv run pytest python/tests` — 1397 passed, 9 skipped
- [x] `uv run black python --check && uv run isort python --check && uv run ruff check python && uv run pyright python` — clean (0 errors)
- [x] `cargo fmt --all --check` — clean
- [x] Multi-layer chaining test (`test_no_return_chains_two_cz_layers_without_returning_home`) verifies layer N's output flows into layer N+1 with no palindrome return and accumulated `move_count` reflects only forward moves.
- [x] `rust_nodes_expanded_total` accumulator tests use an unaligned initial state so the strict `>` assertion is falsifiable.
- [x] Codecov patch coverage above the 88.71% threshold (88% → ≥89% after edge-path tests for `_no_return_base.py` and a Rust unit test for `is_home_position`).

## BREAKING CHANGE — API surfaces affected

Marked with the `breaking` label. Surfaces enumerated for downstream impact audit:

**Python API**:
- `solve_options_from_traversal(...)` return type changed (now `tuple[SolveOptions, EntropyOptions]` instead of a single `SolveOptions`)
- `SolveResult.move_layers` / `MultiSolveResult.move_layers` element type changed (raw 6-tuples → `LaneAddress` objects); resolves #564
- `EntropyScorer.metrics` / `score_moveset` / `apply_moveset` argument type changed (typed `LaneAddress` instead of raw tuples); resolves #564
- `NoReturnPlacementStrategy.strategy` (and `deadlock_policy`) take `SearchStrategy` / `DeadlockPolicy` enums, not strings (no prior release — only "breaking" for prereleased uses of this branch)

**Python typing stubs** (`_native.pyi`):
- `SolveOptions` constructor signature changed (fields moved to new `EntropyOptions`)
- New option types: `EntropyOptions`, `EntanglingOptions`, `NoHomeOptions`, `RecedingHorizonOptions`
- New `MoveSolver` methods: `solve_entangling`, `solve_nohome`, `solve_entangling_rh`

**Rust crate API**:
- `SolveOptions` struct fields removed (split into `SolveOptions`/`EntropyOptions`/`EntanglingOptions`)
- `solve()` signature extended (now takes an optional `EntropyOptions` argument)
- `SearchContext` gained a `cz_pairs: Option<&[(u32, u32)]>` field
- `MOVE_PENALTY`, `LOOKAHEAD_BETA`, `OCCUPANCY_PENALTY_DEFAULT` moved from `solve.rs` to `entangling.rs` (visible as `pub(crate)`)

**C API**: unaffected.

## Follow-ups (not in this PR)

- Split `entangling.rs` (~2.5k LOC) into submodules if reviewers prefer. Main's `entropy.rs` is similarly sized, so no current convention pressure.
- Promote the implicit `cz_pairs.is_some()` "loose-goal mode" check to an explicit `SearchMode { FixedTarget, NoReturn { cz_pairs } }` Rust enum.
- Replace the removed `BLOQADE_RH_PROFILE` env-var telemetry with `tracing`-crate based structured tracing if RH profiling is needed again.
- Surface `rust_nodes_expanded_total` through `PalindromePlacementStrategy` (pre-existing gap on main).
- Refactor the three new `MoveSolver.solve_*` PyO3 methods to a builder/options-struct approach to drop the `#[allow(clippy::too_many_arguments)]` annotations.


python/bloqade/lanes/heuristics/physical/movement.py
```diff
@@ -59,61 +45,56 @@
 ]
 
 
-_DIR_MAP = {0: Direction.FORWARD, 1: Direction.BACKWARD}
-_MT_MAP = {0: MoveType.SITE, 1: MoveType.WORD, 2: MoveType.ZONE}
-
-
 def convert_move_layers(
```
weinbe58, line 48: is this function being used anywhere? If not we can just remove it.

weinbe58 (commented):
@lsschmid — solid PR overall. The architecture is clean, backward compatibility is well-reasoned, and the design document is thorough. A few things to address before merge:

---

## Must / Should Fix

### 1. `DistanceTable` flat index — use a 2D array, not a `HashMap`

The new `flat_distance: HashMap<(u64, u64), u32>` effectively doubles distance storage. The arch spec gives a fixed, small number of sites (~100), so the right structure is a 2D array — a flat `Vec<u32>` with stride `n_sites` (~10,000 entries, ~40 KB) indexed by compact location IDs. A `HashMap<u64, usize>` side index built once at construction is all you need to map encoded locations to array indices. The array layout also gives better cache behavior on the Hungarian cost-matrix build (`entangling.rs` pair × slot loop), which is the hot path this optimization is meant to serve.

### 2. `plans/` — wrong directory

`plans/2026-05-11-receding-horizon-loose-goal-design.md` belongs at `docs/superpowers/plan/2026-05-11-receding-horizon-loose-goal-design.md`. The project convention is `docs/superpowers/spec/` for specs and `docs/superpowers/plan/` for design/plan documents. A bare `plans/` at the repo root is not the convention.

### 3. Hardcoded strategy strings → use the `_native` enums

`NoReturnStrategyBase.strategy: str = "ids"`, `deadlock_policy: str = "move_blockers"`, and the `_STRATEGY_MAP` / `_DEADLOCK_MAP` dispatch tables mean users get a `KeyError` at runtime for a typo. `_native.SearchStrategy` and `_native.DeadlockPolicy` are already exported Python enums — the `strategy` and `deadlock_policy` fields on all three strategy classes should accept the enum type directly (keep the string shim internally for backward compat if needed, but make the enum the primary interface). This catches misspellings at construction time.

### 4. Add a multi-layer chaining test

All new Python tests call `cz_placements` once. The key contract of no-return placement is that the output layout of layer N feeds directly into layer N+1 without palindrome returns. Please add at least one test that chains two CZ layers and verifies the accumulated `move_count` reflects only forward moves.

---

## Should Fix

### 5. Breaking change footer needs API surface categorization

The `breaking` label is correct. The footer should enumerate the surfaces explicitly so downstream users can audit impact:

- **Python API**: `solve_options_from_traversal` return type changed; `SolveResult.move_layers` / `MultiSolveResult.move_layers` type changed; `EntropyScorer.metrics` / `score_moveset` argument type changed
- **Python typing stubs** (`_native.pyi`): `SolveOptions` constructor signature changed (fields moved to `EntropyOptions`); three new option types added; new `MoveSolver.solve_*` methods
- **Rust crate API**: `SolveOptions` struct fields removed; `solve()` signature extended; `SearchContext` gained `cz_pairs` field
- **C API**: unaffected (confirm)

### 6. `validate_initial_layout` no-op — confirm intent

```python
def validate_initial_layout(self, initial_layout: ...) -> None:
    _ = initial_layout
```

If `PlacementStrategyABC.validate_initial_layout` is already a no-op in the base, this override is redundant and should be deleted. If it's not, this silently removes base validation for all no-return strategies — confirm that's intentional and add a comment explaining why.

### 7. `test_..._exposes_rust_nodes_expanded` is unfalsifiable for the current test state

`assert strategy.rust_nodes_expanded_total >= before` always passes when `nodes_expanded=0`. Use an initial state that actually requires at least one expansion (qubits not yet at a valid entangling configuration) so the `>` assertion holds. The smoke test already constructs such a state — reuse it.

---

## Nits

- `LaneIndex::Clone` comment says "so callers can wrap in an `Arc`" — verify that `LooseTargetGenerator` actually stores `Arc<LaneIndex>` and not a cloned copy; otherwise the clone happens K times per receding-horizon stage.
- `LOOKAHEAD_BETA`, `MOVE_PENALTY`, `OCCUPANCY_PENALTY_DEFAULT` are defined at the top of `solve.rs` but used exclusively by entangling-path code — they'd be easier to find in `entangling.rs`.
- `_STRATEGY_MAP` has `"cascade"` and `"cascade-ids"` mapping to the same value — a brief comment marking `"cascade"` as an alias prevents confusion.
- `#[allow(clippy::too_many_arguments)]` on three new `solve_*` PyO3 methods is fine for now but worth tracking; a builder/options-struct approach would clean this up eventually.

weinbe58:
One more question for you @lsschmid — on the `_native.pyi` changes: some of these appear to be bundled cleanup rather than required by the no-return feature itself. Would you be able to separate them to reduce the breaking surface of this PR?

Specifically, the changes break down into three buckets:

**Required for this PR:**
- `SolveOptions` split into `SolveOptions` + `EntropyOptions` — necessary because the new `solve_entangling`/`solve_nohome`/`solve_entangling_rh` entry points have no use for `max_movesets_per_group`, `max_goal_candidates`, etc., and shouldn't require callers to pass entropy-specific knobs.

**Bundled cleanup (not required):**
- `SolveResult.move_layers` / `MultiSolveResult.move_layers` type change (raw 6-tuples → `LaneAddress` objects): All three new entry points share the same `PySolveResult` type as the existing `solve()`. `_no_return_base.py` could unpack raw tuples exactly as `convert_move_layers` did before, without touching the existing `solve()` path at all.
- `EntropyScorer.metrics` / `score_moveset` / `apply_moveset` signature change: `EntropyScorer` is only used in the visualization path — none of the new placement strategies touch it.

The `LaneAddress`-returning `move_layers` is a nicer API and the cleanup is worthwhile, but it expands the set of callers broken by this PR. If it can be moved to a follow-up with no feature dependency, that would keep this PR's breaking footprint to `SolveOptions`-only and make the changelog easier to communicate.

If the cleanup is already done and tested, keeping it here is also reasonable — just want to make sure it's a deliberate choice rather than an accidental scope expansion.

weinbe58:
Correction to my previous comment — the `move_layers` tuple → `LaneAddress` change and the `EntropyScorer` signature changes are actually resolving #564 (assigned to @lsschmid), not incidental cleanup. Ignore the suggestion to defer those; they belong here. The only part of that comment that stands is the question of whether to keep both in one PR or split them — but that's a scope call, not a correctness concern.

weinbe58 (changes requested):
See my PR comments.

lsschmid:
@weinbe58 thanks for the thorough review — addressed all 11 items in commit `3df33e5`. Brief walkthrough:

### Must / Should Fix

- **M1 — `DistanceTable` 2D array.** Replaced the `HashMap<(u64, u64), u32>` with a `Vec<u32>` of size `n_loc²` plus a side `HashMap<u64, usize>` mapping each encoded location to a compact index. Also dropped the redundant `distance_to` nested HashMap (the column scan in `for_each_source` covers it). Same access patterns, less memory, better cache locality on the Hungarian hot path.
- **M2 — `plans/` → `docs/superpowers/plans/`.** `git mv`-ed the design doc and updated three code references that pointed at the old path.
- **M3 — Strategy enum-only API.** `NoReturnStrategyBase.strategy` is now `SearchStrategy` (default `IDS`); `deadlock_policy` is now `DeadlockPolicy` (default `MOVE_BLOCKERS`). Typos now fail at construction. `_DEADLOCK_MAP` is gone; `_STRATEGY_MAP` stays for `PhysicalPlacementStrategy`'s legacy traversal API.
- **M4 — Multi-layer chaining test.** Added `test_no_return_chains_two_cz_layers_without_returning_home`: puts two qubits on non-CZ-partner words, runs CZ layer 1 (forces movement), feeds the output into a second CZ layer (already valid → no movement), asserts the accumulated `move_count` equals layer 1's — directly proves the no-return contract.

### Should Fix

- **S5 — Breaking-change footer.** Added to the PR description above with all four API surfaces broken out (Python API, `_native.pyi`, Rust crate, and C API — confirmed unaffected).
- **S6 — `validate_initial_layout` no-op intent.** Added a comment explaining the no-op is intentional: `PlacementStrategyABC.validate_initial_layout` is `@abc.abstractmethod`, so the override is required, and the no-op matches the same pattern `PhysicalPlacementStrategy` already uses.
- **S7 — `rust_nodes_expanded` test now falsifiable.** Tests now use an unaligned initial state (qubits at `(0, 0)` and `(3, 0)` on non-CZ-partner words), so the solver must expand at least one node. The strict `>` assertion now holds; the previous `>=` was unfalsifiable when `nodes_expanded == 0`.

### Nits

- **N8 — `Arc<LaneIndex>` already.** Verified `LooseTargetGenerator.index` is already `Arc<LaneIndex>`. The `index.clone()` calls in receding-horizon stages are `Arc::clone` refcount bumps, not full LaneIndex clones. No code change needed.
- **N9 — Constants relocated.** `MOVE_PENALTY`, `LOOKAHEAD_BETA`, `OCCUPANCY_PENALTY_DEFAULT` moved from `solve.rs` to `entangling.rs` as `pub(crate)`. `receding_horizon.rs` now imports them from `entangling` too instead of redefining locally.
- **N10 — `"cascade"` alias comment.** Added inline in `_STRATEGY_MAP`.
- **N11 — `clippy::too_many_arguments` tracked.** Added to the Follow-ups section of the PR description (the builder/options-struct refactor is the natural fix later).

### Inline

- **`movement.py:48` (`convert_move_layers`)** — Was being duplicated inline in `NoReturnStrategyBase.cz_placements`. Now uses the existing helper at all three call sites (`movement.py`, `move_synthesis.py`, `_no_return_base.py`).

---

Verified end-to-end: `cargo test --workspace`, `uv run pytest python/tests` (1397 passed, 9 skipped), black / isort / ruff / pyright / cargo fmt all clean.

weinbe58 (approved):
LGTM

PR #596, opened 2026-05-01
ci(benchmark): add workflow to run benchmark harness on PRs

### LINKED ISSUE #501: ci(benchmark): add workflow to run benchmark harness
## Summary

Add an automated workflow that executes the benchmark harness so benchmark coverage is reproducible in CI.

## Scope

- Create a GitHub Actions workflow that runs the benchmark harness entrypoint
- Configure environment/setup needed for benchmark execution
- Define trigger strategy (manual and/or scheduled, plus pull request policy as appropriate)
- Persist benchmark outputs as workflow artifacts for inspection

## Parent

Part of #409

Jason-Ludmir:
## Summary

- Add a `harness` matrix job to `.github/workflows/ci.yml` that runs `just benchmark-logical` and `just benchmark-physical` on every PR.
- Each invocation compares against the committed baselines in `python/benchmarks/harness/latest_{logical,physical}.csv` and exits non-zero on drift in algorithm-output fields (move counts, estimated fidelity, nodes explored, max depth, generator id, backend, notes). Wall time is excluded so the check is deterministic.
- The produced CSV is uploaded as an artifact even when the comparison fails, so reviewers can inspect what the run actually produced.
- Update `python/benchmarks/cli.py` to skip kernels with non-Clifford rotations past an initial one (currently `adder_4` and `qpe_9`) from the logical benchmark, since the logical pipeline cannot lower them today; covered by a new test in `python/tests/benchmarks/test_cli.py`.

When a contributor intentionally changes a search algorithm, they regenerate the baseline CSV(s) locally and commit the new file in the same PR — the diff in the PR is the review artifact.

Closes #501

## Test plan

- [ ] CI runs `harness (logical)` and `harness (physical)` on this PR and both pass against the refreshed committed baselines.
- [ ] Both jobs upload their `benchmark-{logical,physical}-csv` artifacts.
- [ ] Verify (in a follow-up sanity check, not this PR) that touching `python/bloqade/lanes/heuristics/move_synthesis.py` in a way that perturbs search output does cause the workflow to fail.


.github/workflows/ci.yml
```diff
@@ -138,6 +138,43 @@ jobs:
       - name: Run demos
         run: uv run just demo
 
+  harness:
+    name: harness (${{ matrix.architecture }})
```
weinbe58:
```suggestion
  benchmarks:
    name: benchmarks (${{ matrix.architecture }})
```

just rename the workflow

weinbe58 (changes requested):

.github/workflows/ci.yml
Jason-Ludmir: Fixed.

weinbe58 (approved):
LGTM

PR #598, opened 2026-05-01
feat(rewrite): place-stage ASAP reorder — MergeStaticPlacement, ReorderStaticPlacement, SplitStaticPlacement

weinbe58:
## Summary

Three composable rewrite passes for the \`place\`-dialect compilation stage, each with a pluggable policy:

1. **\`MergeStaticPlacement(merge_policy)\`** — merges consecutive \`StaticPlacement\` nodes; replaces \`MergePlacementRegions\`
2. **\`ReorderStaticPlacement(reorder_policy)\`** — ASAP-schedules gate statements within a merged body via \`rustworkx.PyDAG\`
3. **\`SplitStaticPlacement(split_policy)\`** — re-splits into CZ-anchored layers

Concrete policies: \`gate_only_merge\`, \`sq_only_merge\`, \`asap_reorder_policy\`, \`cz_layer_split_policy\`.

Two pluggable passes exposed in \`bloqade.lanes.passes\`:
- **\`SequentialPlacePass\`** (default) — merges only SQ-gate placements, preserves program order
- **\`ASAPPlacePass\`** — full ASAP optimization; strips debug statements, merges all gate placements, reorders, fuses, re-splits

Full design: \`docs/superpowers/specs/2026-05-01-place-asap-reorder-design.md\`

## Progress

- [x] Task 1 — \`MergeStaticPlacement\` + \`gate_only_merge\`
- [x] Task 2 — \`ReorderStaticPlacement\` + \`asap_reorder_policy\`
- [x] Task 3 — \`SplitStaticPlacement\` + \`cz_layer_split_policy\`
- [x] Task 4 — Wire into pipeline; remove \`merge_heuristic\` param

## Breaking changes (Python API)

The following public symbols are removed with no deprecation path:

- \`upstream.default_merge_heuristic\`
- \`upstream.always_merge_heuristic\`
- \`merge_heuristic\` parameter on \`LogicalPipeline\`, \`PhysicalPipeline\`, \`NativeToPlace\`, \`squin_to_move\`, and the four \`metrics.py\` compile helpers

Migration: replace \`merge_heuristic=always_merge_heuristic\` with \`place_opt_type=ASAPPlacePass\`.

## Test plan

- [x] \`uv run pytest python/tests/ -m "not slow" -v\` — 1084 pass
- [x] \`uv run pyright python/\` — 0 errors
- [x] Slow integration test: \`test_asap_place_pass_roundtrip_state_vector\` (GHZ circuit, state-vector equivalence check)


python/bloqade/lanes/rewrite/split_static_placement.py
```diff
@@ -0,0 +1,106 @@
...
+    """
+    old_yield = body_block.last_stmt
+    assert isinstance(old_yield, place.Yield)
+    classical_results = tuple(old_yield.classical_results)
+
+    stmts: list[_GateStmt] = [
+        s
+        for s in body_block.stmts
+        if isinstance(
+            s, (place.R, place.Rz, place.CZ, place.Initialize, place.EndMeasure)
```
weinbe58: why is `EndMeasure` here?

python/bloqade/lanes/rewrite/split_static_placement.py
```diff
@@ -0,0 +1,106 @@
...
+from kirin import ir
+from kirin.rewrite import abc
+
+from bloqade.lanes.dialects import place
+from bloqade.lanes.types import StateType
+
+_GateStmt = place.R | place.Rz | place.CZ | place.Initialize | place.EndMeasure
+
+
+def cz_layer_split_policy(body_block: ir.Block) -> list[ir.Block]:
```
weinbe58, line 13: Ah we're missing which qubits are participating within each of the blocks I think the signature should be:

`Callable[[ir.Block], list[tuple[ir.Block, list[int]]]`

and then use the second element to indicate which qubit ssa values to pass into the split up `StaticPlacement`.

python/bloqade/lanes/rewrite/split_static_placement.py
```diff
@@ -0,0 +1,130 @@
...
+                new_stmt = stmt.from_stmt(
+                    stmt,
+                    attributes={
+                        "qubits": ir.PyAttr(tuple(remap[q] for q in stmt.qubits))
+                    },
+                )
+                new_stmt.insert_before(stmt)
+                for old_r, new_r in zip(stmt.results, new_stmt.results, strict=True):
+                    old_r.replace_by(new_r)
+                stmt.delete()
```
weinbe58: I'm pretty sure you can replace this code with`stmt.replace_by(new_stmt)`.

weinbe58:
## Review

Overall this is solid work. The pass design is clean, the pluggable-policy pattern is consistent across all three passes, and the unit test coverage is thorough. A few things need attention before merging.

---

### 1. PR description is stale — all 4 tasks are shipped, not just Task 1

The checklist says Tasks 2–4 are incomplete, but the diff includes all of them:
- `reorder_static_placement.py` — Task 2 ✓
- `split_static_placement.py` — Task 3 ✓
- `passes.py` + pipeline wiring + `merge_heuristic` removal — Task 4 ✓

Please update the description and check all boxes.

---

### 2. Breaking API changes need the `breaking` label

These public symbols are removed with no deprecation path:

- `upstream.default_merge_heuristic`
- `upstream.always_merge_heuristic`
- `merge_heuristic` parameter on `LogicalPipeline`, `PhysicalPipeline`, `NativeToPlace`, `squin_to_move`, and all four `metrics.py` methods

Per AGENT.md: tag with the `breaking` label and add `BREAKING CHANGE:` footer to the commit. Callers using `merge_heuristic=...` will get a `TypeError` at runtime with no indication of the migration path.

---

### 3. `HoistConstants` runs twice in `upstream.NativeToPlace`

`upstream.py` retains the explicit `HoistConstants` call and then delegates to `place_opt_type`, but `SequentialPlacePass.unsafe_run` starts with its own `HoistConstants`. The base-class path (`_NativeToPlaceBase.emit`) correctly removes the standalone call and relies entirely on the pass. The `upstream.NativeToPlace` path doesn't — the redundant call should be removed there too:

```python
# upstream.py NativeToPlace.emit — remove this line:
rewrite.Walk(circuit2place.HoistConstants()).rewrite(out.code)
# ↑ now duplicated by SequentialPlacePass/ASAPPlacePass
self.place_opt_type(out.dialects, no_raise=no_raise)(out)
```

Running it twice is idempotent but inconsistent and wasteful.

---

### 4. `_GateStmt` alias means different things in the two new files

`reorder_static_placement.py`:
```python
_GateStmt = place.R | place.Rz | place.CZ | place.Initialize | place.EndMeasure
```

`split_static_placement.py`:
```python
_GateStmt = place.R | place.Rz | place.CZ
```

Same name, different types. Pick distinct names (`_GateOrBarrierStmt` vs `_GateStmt`) or consolidate in a shared types module.

---

### 5. `RemoveDebugStatements` only runs in `ASAPPlacePass` — undocumented behavioral difference

Users who switch from `SequentialPlacePass` (default) to `ASAPPlacePass` silently lose `debug.Info` statements. If that's intentional (ASAP can't schedule through opaque debug nodes), the docstring should say so.

---

### 6. No end-to-end integration test for `ASAPPlacePass`

`test_default_and_always_merge_have_same_operations` was the only test that verified the full pipeline (circuit → move → Stim) for a merge-policy variant; it's deleted here. There's no equivalent test confirming that `ASAPPlacePass` produces valid, equivalent Stim output compared to `SequentialPlacePass`. Given that this touches the core compilation path, an integration test (even marked `@pytest.mark.slow`) would give confidence that the new pass sequence is sound end-to-end.

---

### Minor

- `always_merge` is exported from `circuit2place.py` and imported in the test, but it's not exported from `heuristics/physical/__init__.py` or `passes.py`. If it's meant to be a public policy for users, add it to `__all__`. If it's test-only, prefix with `_`.
- `_asap_schedule`: `node_for[i] = dag.add_node(i)` — rustworkx guarantees sequential node indices when nothing is deleted, so this works, but it might be clearer to just use `i` directly as the node index and remove `node_for`. Low priority.

---

### What's correct

- The merge logic in `MergeStaticPlacement` is a faithful port of `MergePlacementRegions` with the correct policy-signature upgrade. The SSA result-replacement logic (`strict=True` zip, accumulating classical results) is correct.
- The ASAP DAG construction and layer assignment are correct.
- `cz_layer_split_policy` returns the original block (no-op) when there's ≤1 CZ, preventing unnecessary splitting.
- `sq_only_merge` is semantically equivalent to the old `default_merge_heuristic` — behavior is preserved for the default path.
- `FuseAdjacentGates` fires after reorder but before split, which is the correct position per the design doc.

weinbe58:
@jasonhan3 Can you try out this branch and see how the gadgets are working? you don't need to review the code.

Here is how you invoke the compiler with the ASAP optimization:

```python
from bloqade import squin
from bloqade.gemini.common.dialects import qubit
from bloqade.lanes.passes import ASAPPlacePass
from bloqade.lanes.pipeline import PhysicalPipeline
from bloqade.lanes.visualize import debugger


kernel = squin.kernel.add(qubit)
kernel.run_pass = squin.kernel.run_pass

@kernel
def main():
    ...

move_mt = PhysicalPipeline(place_opt_type=ASAPPlacePass).emit(main)
debugger(move_mt, get_arch_spec())
```

PR #600, opened 2026-05-04
feat: Move Policy & Target Generator DSL framework (#597)

### LINKED ISSUE #597: feat: Move Policy DSL + Target Generator DSL framework
## Overview

This tracking issue covers the full implementation of the **Move Policy DSL framework** described in `docs/superpowers/specs/2026-04-20-search-heuristic-dsl-design.md`. The framework introduces two Starlark-hosted DSLs that let agents author search heuristics without writing Rust, over a shared Rust substrate (`bloqade-lanes-dsl-core`).

**Design document**: `docs/superpowers/specs/2026-04-20-search-heuristic-dsl-design.md`

---

## Architecture

```
Caller (Python / Rust / CLI)
    └─▶ Move Kernel (Rust)         — SearchGraph, AOD validation, budget enforcement
            └─▶ bloqade-lanes-dsl-core — starlark-rust adapter, sandbox, shared primitives
                    └─▶ Policy (.star)  — agent-authored: init(root, ctx), step(graph, gs, ctx, lib)
```

The kernel owns all safety-critical invariants (AOD validation, transposition-table dedup, budget enforcement). The policy owns algorithm-level decisions. Neither inspects the other's internals.

---

## Plan A — Move Policy DSL + Shared Substrate ✅ (PR #TBD / branch `jason/move-policy-dsl`)

**New crate**: `bloqade-lanes-dsl-core`
- `adapter.rs` — starlark-rust integration (parse, bind, invoke, marshal)
- `sandbox.rs` — resource limits (`starlark_steps_per_tick`, `starlark_memory_per_solve`), forbidden builtins
- `policy_trait.rs` — generic `Policy<Handle, Action, InitArg, NodeState>` trait
- `primitives/` — shared Starlark value types (`LocationAddress`, `LaneAddr`, `Config`, `MoveSet`, `ArchSpec` wrapper, `stable_sort`, `argmax`, `normalize`)
- `errors.rs` — `DslError` variants with Starlark line:col info

**Extended**: `bloqade-lanes-search/src/move_policy_dsl/`
- `kernel.rs` — Move Kernel loop (algorithm-agnostic; drives `step()` calls)
- `actions.rs` — `insert_child`, `update_node_state`, `update_global_state`, `emit_solution`, `halt`, `invoke_builtin`
- `graph_handle.rs` — mediated `PolicyGraph` view exposed to policies
- `lib_move.rs` — `score_lanes`, `top_c_per_qubit`, `group_by_triplet`, `pack_aod_rectangles`, `hop_distance`, `mobility`, etc.
- `adapter_impl.rs` — glue between Move DSL and dsl-core's `Policy` trait

**Extended**: `bloqade-lanes-bytecode-python` — PyO3 `solve(..., policy_path=..., policy_params=...)` binding

**Reference policy**: `policies/reference/entropy.star` — Starlark reproduction of `Strategy::Entropy`

**Python API**:
```python
solver = MoveSolver.from_json(arch_json)
result = solver.solve(
    initial=[(0, loc_a), (1, loc_b)],
    target=[(0, loc_c), (1, loc_d)],
    blocked=[],
    policy_path="policies/reference/entropy.star",
    policy_params={"e_max": 12},
    max_expansions=10_000,
    timeout_s=30.0,
)
```

---

## Plan B — Target Generator DSL 🔲

A Starlark mirror of `TargetGeneratorABC` (from #533). Pure function interface — no state, no tree, no step loop.

**New file**: `python/bloqade/lanes/heuristics/physical/target_generator_dsl.py`
- Adapter implementing `TargetGeneratorABC` by loading a `.star` file via dsl-core (PyO3)
- Wraps `TargetContext` as a Starlark handle with: `ctx.arch_spec`, `ctx.placement`, `ctx.controls`, `ctx.targets`, `ctx.lookahead_cz_layers`, `ctx.cz_stage_index`
- Validates candidates on the Rust side: qid coverage, location group validity, CZ-blockade pair invariant

**Policy shape**:
```python
def generate(ctx):
    # Returns list[dict[int, LocationAddress]]
    ...
```

**Integration**: injected at the `target_generator=` field of `PhysicalPlacementStrategy` — no new orchestration code needed.

---

### Additional reference policies (`policies/reference/`)
- `dfs.star` — vanilla depth-first with LIFO stack in `GlobalState`
- `bfs.star` — breadth-first with a FIFO
- `ids.star` — iterative deepening

Each file is heavily commented; agents are expected to start from one as a template.

### CLI harness (`bloqade-lanes-bytecode-cli`)
```bash
# Run a policy on a fixture
lanes eval-policy --policy path/to/policy.star \
                  --case fixtures/sample.json \
                  --max-expansions 10000

# Verbose per-step trace
lanes trace-policy --policy path/to/policy.star \
                   --case fixtures/sample.json
```
Fixture format: JSON with `arch_spec_path`, `initial`, `target`, `blocked`, optional `params` override.

### Auto-generated primer (`policies/primer.md`)
Documents every `lib.*`, `actions.*`, `graph.*`, and `ctx.*` name with signature and semantics. Regenerated from Rust-side type definitions via a build step; drift is a CI-enforced invariant.

### Snapshot tests
Each reference policy runs against 3–4 fixtures with committed expected `SolveResult` JSON. Regressions caught by CI.

---

## Sandbox constraints (all plans)

- No I/O, network, wall clock, RNG, `load()` of external files
- Per-tick Starlark step budget (default: 1,000,000 steps)
- Per-solve heap cap (default: 128 MiB)
- Deterministic: byte-identical results for same inputs — ties broken on structural id, not hash randomness

## Determinism guarantees

1. No RNG exposed through any `lib.*` primitive
2. All collection-returning primitives return sorted or insertion-order vecs
3. `stable_sort` / `argmax` use stable ordering on (value, structural id)
4. Hermetic inputs: policies see only `(initial config, target, blocked, PARAMS, ctx)`
5. FNV-1a `Config` hash is deterministic across runs and platforms

## Open questions

- Exact defaults for `starlark_steps_per_tick` / `starlark_memory_per_solve` (determined empirically)
- `graph.last_insert()` side-channel vs. action-as-expression ergonomics — revisit after agent usage
- `PARAMS` override granularity: only top-level fields today; nested struct overrides may be needed

## Out of scope

- Replacing existing Rust search traits or Python `TargetGeneratorABC`
- Production-throughput parity (~10× overhead expected; stabilized policies can be ported to native Rust)
- Hot reload, untrusted remote policies, differentiable/learning-based policies, cross-policy composition
- Unified Move + Place DSL (forward-compat path exists via generic `Policy` trait; see design §11)

Jason-Ludmir:
## Summary

This PR lands the full Starlark-based policy layer for move synthesis and target generation, then wires that DSL into a Roger-luo/autotune loop so LLM-authored `.star` policies can be iterated against real physical-placement benchmarks without recompiling Rust.

Closes #597.

## What changed

- Added the shared `bloqade-lanes-dsl-core` substrate: deterministic Starlark sandboxing, a generic policy trait, shared primitive wrappers, structured errors, and utility bindings used by both DSLs.
- Added the Move Policy DSL under `bloqade-lanes-search/src/move_policy_dsl/`: kernel run loop, action verbs, `graph.*` / `lib.*` / `ctx.*` handles, observer hooks, `PARAMS_OVERRIDE`, timeout/budget support, and Python-facing `MoveSolver.solve(policy_path=..., policy_params=..., timeout_s=...)` plumbing.
- Added the Target Generator DSL under `target_generator_dsl/`: Starlark policies can generate CZ-stage target candidates, with validation delegated to the existing physical-placement target validator and a Python `TargetGeneratorDSL` adapter.
- Added CLI and fixture infrastructure: `bloqade-bytecode eval-policy`, `bloqade-bytecode trace-policy`, NDJSON trace output, shared move/target problem fixture loading, snapshot fixtures, `dsl_snapshot` tests, and generated `policies/primer.md`.
- Added autotune integration: `.autotune.toml`, `policies/autotune/candidate.star`, `scripts/autotune/measure_dsl_policy.py`, `.python-version`, `.autotune/` ignore rules, and a `dsl_autotune` benchmark strategy selected by `BLOQADE_DSL_POLICY`.
- Fixed several issues discovered while running the autotune loop on physical Gemini problems, especially u64 lane encoding round-trips and Python acceptance of valid DSL fallback results.

## Move Policy DSL

The Move Policy DSL lets users write a single `.star` file exporting:

```python
init(root, ctx) -> global_state
step(graph, gs, ctx, lib) -> action | list[action]
```

Policies drive the Rust search kernel through a small action vocabulary:

- `insert_child(parent, move_set_encoded)` creates a search child from an encoded AOD move set.
- `update_node_state(node, patch)` and `update_global_state(patch)` store deterministic Starlark-visible policy state.
- `emit_solution(node)` returns a found path.
- `halt(status, message)` terminates explicitly.
- `invoke_builtin(name, args)` supports controlled builtins such as `sequential_fallback`.

The exposed read surface is deliberately narrow and deterministic. `graph` gives policies access to configs, parents, goal checks, insert outcomes, and per-node state. `ctx` exposes targets, blocked locations, and the architecture. `lib` exposes distance and mobility helpers plus the Rust-backed AOD candidate pipeline.

The autotune work pushed the Move DSL surface further than the initial framework by moving strategy-owned scoring and pruning decisions into Starlark. In particular, policies can now build their own candidate ordering with primitives such as `unresolved_qubits`, `legal_lanes`, `scored_lane`, and `pack_aod_rectangles`, while the Rust side still owns safety-critical move validation and graph insertion.

## Target Generator DSL

The Target Generator DSL mirrors the Move DSL pattern for CZ-stage target selection. A `.star` policy can emit candidate qubit-to-location placements for a CZ stage, and the Rust/Python bridge validates those candidates through the existing physical placement validator.

This gives us two replaceable policy layers:

- target generation: choose where the CZ pair should be physically placed;
- move policy: choose how to move from the current physical state to that target.

The Python adapter is `TargetGeneratorDSL(TargetGeneratorABC)`, so it plugs into `PhysicalPlacementStrategy(target_generator=...)` without changing call sites that already consume target generators.

## Autotune integration

The new autotune path is intended to answer: can an LLM evolve a useful Move Policy DSL search strategy by editing only a `.star` policy file?

The branch adds a Roger-luo/autotune task named `dsl-move-policy-v2-steane35-no-builtin`. Its tunable path is `policies/autotune/**`, while reference policies and design docs are denied so generated policies cannot simply copy the examples. The configured objective hierarchy is strict:

- maximize `success_rate` first;
- minimize `move_layers` only among successful policies;
- emit wall time and nodes explored as diagnostics, not as primary optimization targets.

The current measurement command runs:

```bash
env BLOQADE_DSL_KERNELS=steane_physical_35 MPLCONFIGDIR=.autotune/matplotlib uv run python scripts/autotune/measure_dsl_policy.py
```

The measure script calls `compile_squin_to_move` directly with the physical Gemini arch spec and a `PhysicalPlacementStrategy` whose `RustPlacementTraversal` points at the policy under evaluation. It intentionally bypasses the benchmark CLI's architecture-mode plumbing because that path mutates logical-initialization settings that are unrelated to physical-arch move search.

Metrics are emitted as `AUTOTUNE_METRIC <name> <value>` lines for a RegexAdaptor. Per-case diagnostics and the policy file contents are written to stderr so each iteration preserves enough evidence to debug failures after temporary worktrees are discarded.

## Current autotuned candidate

`policies/autotune/candidate.star` is currently a no-builtin greedy best-first search for `steane_physical_35` on the physical Gemini arch.

Important policy properties:

- It uses a frontier of `[priority, node_id]` entries and pops the minimum priority.
- The priority is pure remaining hop distance `h`, summed only over unresolved qubits.
- It enumerates legal lanes only for unresolved qubits, preventing the packer from dragging already-at-target qubits away from their goals.
- It scores lanes as `(d_cur - d_dst) * d_cur + d_cur`, which rewards forward progress for far-away qubits and gives lateral moves for far-away qubits a positive contribution instead of making them look useless.
- It passes every Rust-packed AOD rectangle through; there are no branch caps, beam widths, top-k slices, per-qubit top-c pruning, fallback halts, or builtin invocations.
- It uses an `awaiting -> queue -> frontier-pop` state machine so each `insert_child` is observed via `graph.last_insert()` before the child is admitted into the frontier.

This candidate is the result of several autotune iterations. Earlier attempts exposed real integration bugs and policy-shaping problems: a fallback-first policy could technically solve by delegating to Rust, reference-policy copies were too brittle, and branch caps/top-c pruning hid search completeness failures. The current task prompt explicitly forbids those shortcuts so the LLM has to express the search behavior in Starlark.

## Bugs found by autotune and fixed here

The autotune loop acted as a stress test for the DSL on real physical Gemini lane IDs. It found several issues that were not obvious from the original framework tests:

- Starlark int to JSON marshalling truncated physical Gemini lane encodings above `i32::MAX`, turning valid lanes into `0` and causing `aod_invalid: lane 0 has no qubit at src`.
- `insert_child` action conversion had the same big-int truncation class when converting Starlark lists into encoded move sets.
- `StarlarkLane.encoded` cast `LaneAddr::encode_u64()` to `i32`, which broke policies that used `lane.encoded` directly.
- The Python physical-placement bridge rejected valid DSL results produced through the documented `sequential_fallback` path because those results carry `policy_status="fallback: ..."` rather than `status="solved"`, despite having non-empty move layers and a target goal config.
- The measurement loop initially discarded benchmark failure notes; the script now preserves policy contents and per-case notes in iteration stderr so the next research pass can reason about actual failures.

## User-facing surfaces

Python:

```python
from bloqade.lanes.bytecode import MoveSolver

solver = MoveSolver.from_json(arch_json)
result = solver.solve(
    initial=[(0, loc_a), (1, loc_b)],
    target=[(0, loc_c), (1, loc_d)],
    blocked=[],
    policy_path="policies/autotune/candidate.star",
    policy_params={},
    max_expansions=10_000,
    timeout_s=30.0,
)
```

Physical placement:

```python
PhysicalPlacementStrategy(
    traversal=RustPlacementTraversal(
        policy_path="policies/autotune/candidate.star",
        policy_params={},
        timeout_s=30.0,
    )
)
```

Benchmark/autotune:

```bash
BLOQADE_DSL_POLICY=policies/autotune/candidate.star \
BLOQADE_DSL_KERNELS=steane_physical_35 \
uv run python scripts/autotune/measure_dsl_policy.py
```

CLI:

```bash
bloqade-bytecode eval-policy --policy policies/reference/entropy.star --problem policies/fixtures/move/small/problem.json
bloqade-bytecode trace-policy --json --policy policies/reference/entropy.star --problem policies/fixtures/move/small/problem.json
```

## Test plan

Previously verified on this branch:

- [x] `just test-rust` - Rust unit + integration tests, including `dsl_snapshot`, `primer_golden`, and `cli_policy`.
- [x] `just develop-python && uv run pytest python/tests` - Python suite passed at the time of the Plan C update.
- [x] `just lint` - formatting, header check, primer staleness check, and clippy checks were clean at the time of the Plan C update.
- [x] `cargo clippy -p bloqade-lanes-search --all-targets -- -D warnings`.
- [x] `cargo clippy -p bloqade-lanes-dsl-core --all-targets -- -D warnings`.
- [x] CLI smoke: `bloqade-bytecode eval-policy --policy policies/reference/entropy.star --problem policies/fixtures/move/small/problem.json`.
- [x] CLI smoke: `bloqade-bytecode trace-policy --json ...` produced valid NDJSON.
- [x] `just check-primer`.

Latest branch replacement/update verification:

- [x] Pushed `jason/move-policy-dsl` to `030ece300cc7c0d4cdd2657c7bab3879170eba1e`.
- [x] Verified local `jason/move-policy-dsl`, `origin/jason/move-policy-dsl`, and the current autotune worktree branch all point to `030ece300cc7c0d4cdd2657c7bab3879170eba1e`.
- [x] Commit hooks passed when adding the latest policy comments and design notes: end-of-file, trailing whitespace, pyright, and skipped non-applicable format/lint hooks.
- [x] Cursor lints reported no diagnostics for `policies/autotune/candidate.star`.

Not rerun after the final branch replacement: the full Rust/Python test suite. The latest post-replacement change was documentation plus comments in `policies/autotune/candidate.star`.

## Review notes

- This PR is intentionally large because the DSL substrate, Move DSL, Target DSL, CLI harness, fixtures, and autotune integration are coupled: autotune needs the Python physical-placement bridge and the Move DSL primitives, while the DSL framework needs CLI/fixture coverage to be reviewable.
- The autotune task currently focuses on `steane_physical_35`, where native strategies show enough event-count spread to produce meaningful tuning signal.
- `.autotune/` task state is ignored. The committed artifacts are the reproducible config, measure script, candidate policy, and documentation, not per-iteration ledgers/worktrees.
- The CLI binary now links `starlark-rust` transitively. The C FFI library is unaffected because it does not link the policy execution paths.
- PR is still draft until reviewers are comfortable with the combined DSL and autotune scope.


crates/bloqade-lanes-dsl-core/src/policy_trait.rs
```diff
@@ -0,0 +1,43 @@
...
+    Solved,
+    Unsolvable,
+    BudgetExhausted,
+    Fallback(String),
+    Error(String),
+}
+
+/// A hosted algorithm. Type parameters keep this generic enough to host
+/// future placement DSLs (§11 of the spec) without modification.
+pub trait Policy {
```
weinbe58: The `Policy` trait is exported as `pub` but neither `MovePolicy` (Move DSL kernel) nor `TargetPolicy` (Target DSL kernel) actually implement it — they construct their own private structs that satisfy the same shape without the trait bound. Similarly `StepResult<A>` and `HaltStatus` are pub but never pattern-matched by external consumers.

Was this trait intended to be a future constraint — something kernels will eventually be required to implement once there are multiple policy types co-existing — or is it documentation-only scaffolding at this point? If it's the former, it might be worth adding a comment that says so (or actually wiring it up). If it's the latter, making it `pub(crate)` would keep the intent clearer and reduce the exported surface.

weinbe58 (commented):
Architectural review of `bloqade-lanes-dsl-core` — inline comments below on the specific findings. Overall the implementation is clean: no panics, no unsafe, no unwrap in production paths, and all types that matter are wired and tested. The questions are mostly about intent and long-term direction rather than bugs.

crates/bloqade-lanes-dsl-core/src/primitives/types.rs
```diff
@@ -0,0 +1,409 @@
...
+        match attr {
+            // Physical-Gemini lane encodings routinely exceed `i32::MAX` (the
+            // `LaneAddr::encode_u64` packs zone/bus/site bits into the high
+            // half of a `u64`). Casting to `i32` here silently wraps the
+            // value — typically to `0` for the high-bit-only encodings —
+            // and the resulting `lanes: [0]` makes every `insert_child`
+            // call fail with `aod_invalid: lane 0 has no qubit at src`,
+            // silently corrupting every DSL search policy that reads
+            // `lane.encoded` to forward to `insert_child`.
+            "encoded" => Some(heap.alloc(self.0.encode_u64() as i64)),
```
weinbe58, line 117: This cast is documented, but the failure mode when it wraps is pretty far from this line — the policy author would see `insert_child` reject a child with a cryptic error rather than an error that points here.

Would it be worth adding a debug-mode assertion (`debug_assert!(self.0.encode_u64() <= i64::MAX as u64)`) to catch this in tests, or surfacing an explicit `anyhow::bail!` here if the high bit is set? The comment on lines 97-104 is good — just wondering if there's a cheaper way to make the error actionable.

crates/bloqade-lanes-search/src/graph.rs
```diff
@@ -8,67 +8,19 @@
...
-    /// Return the encoded lane values (sorted, deduplicated).
-    pub fn encoded_lanes(&self) -> &[u64] {
-        &self.lanes
-    }
-}
+// `MoveSet` was moved to `bloqade_lanes_dsl_core::primitives::move_set` so
+// the dsl-core `ArchSpec` methods can return a validated `StarlarkMoveSet`
+// without dsl-core needing to depend on this crate. Re-exported here so
+// existing `crate::graph::MoveSet` imports across this crate keep working.
+pub use bloqade_lanes_dsl_core::primitives::move_set::MoveSet;
```
weinbe58, line 23: This re-export exists because `MoveSet` was relocated from `search/graph.rs` to `dsl-core` to break a circular dependency (so `arch_spec.move_set()` can return a validated `StarlarkMoveSet` without `dsl-core` depending on `search`). The comment explains this well.

Longer-term question: is the right home for `MoveSet` and `StarlarkMoveSet` in `dsl-core`, or does this feel like it wants a third crate (`dsl-primitives` or similar) that both `dsl-core` and `search` depend on? The re-export preserves the old import path, but callers in `search` still reach into `dsl-core` for a type that conceptually lives in the search domain. No action needed now — just flagging this for the next time the crate boundaries get revisited.

weinbe58:
**Emerging pattern: ad-hoc error wrapper structs** (4 sites, ready to abstract)

Four locations define a trivial 3-4 line struct whose sole job is wrapping a `String` to satisfy `Into<anyhow::Error>` required by `starlark::Error::new_other()`:

| Location | Type name |
|----------|-----------|
| `primitives/types.rs:129-133` | `BadDiscriminant` |
| `primitives/arch_spec.rs:315-321` | `ArchValidationError` |
| `primitives/utilities.rs:86-96` | `Msg` |
| `primitives/arch_spec.rs:116-127` | inline `Msg` |

They're all structurally identical — `#[derive(Debug)]`, `impl Display`, `impl std::error::Error`. A single `pub(crate) struct DslErrorMessage(pub String)` in `errors.rs` would eliminate all four.

Not urgent, but since all four instances have been stable for 90+ days this is probably the easiest cleanup available. Is there a reason the per-site names are load-bearing (e.g. for backtrace readability), or is consolidation fair game?

---

**`StarlarkArchSpec` constructor friction**

Consumers in `search/move_policy_dsl/kernel.rs:220` and `search/move_policy_dsl/kernel.rs:336-347` manually write `StarlarkArchSpec(Arc::new(...))` and wrap `StarlarkLocation` values because `dsl-core` exposes the types but not constructors for them. A `StarlarkArchSpec::new(spec: Arc<ArchSpec>) -> Self` would remove this — callers currently have to import `Arc` on `dsl-core`'s behalf. Minor, but worth noting if there's a cleanup pass coming.

crates/bloqade-lanes-dsl-core/src/policy_trait.rs
weinbe58: Resolved by 3bbec55. Demoted `Policy`, `StepResult<A>`, `HaltStatus`, and the `policy_trait` module to `pub(crate)` and dropped the corresponding `pub use` from `dsl-core/src/lib.rs`. Added `#![allow(dead_code)]` plus a module-doc note explaining the lifecycle — promote back to `pub` (and wire `MovePolicy` / `TargetPolicy` to actually implement it) once there is a second in-crate kernel or an external consumer.

crates/bloqade-lanes-dsl-core/src/primitives/types.rs
weinbe58: Resolved by 9bfb7c4. Skipped the `debug_assert!(encode_u64() <= i64::MAX as u64)` — looked closer at `LaneAddr::encode` in `bytecode-core` and bit 63 of `encode_u64` is the `direction` bit (`(direction as u32) << 31`, then shifted into the upper word of the u64). Every `Direction::Backward` lane therefore legitimately exceeds `i64::MAX`, and the assertion would fire on the common path. The `as i64` cast is bit-preserving but not range-preserving, by design.

What landed instead:

- Strengthened the inline comment to spell out the bit-63 mechanism and the round-trip contract `Starlark i64 → Rust i64 → as u64 → LaneAddr::decode_u64`.
- Added `lane_encoded_attr_preserves_u64_bit_pattern_through_i64` (in `types.rs` tests) which exercises a Backward `Lane(...)` through Starlark and asserts (a) the value surfaces as the expected negative i64, (b) `as u64` recovers the original `encode_u64()` bits, (c) `decode_u64` rebuilds the original lane bit-exactly.

If anyone later adds a `<= i64::MAX` defensive assertion at the cast site, this test will fail and point them at the comment.

crates/bloqade-lanes-search/src/graph.rs
weinbe58: No code change in this PR. The crate-boundaries question (whether `MoveSet` / `StarlarkMoveSet` belong in `dsl-core`, in `search`, or in a third `dsl-primitives` crate that both depend on) is captured in the 2026-05-31 `bloqade-lanes-search` crate review under §6 "Proposed Refactor Direction" (`docs/superpowers/reviews/2026-05-31-bloqade-lanes-search-review.md`, currently on `phil/refactor-search-infra`). The re-export at `graph.rs:23` stays as-is for this PR and will be revisited when the search-crate refactor reorganizes the type layout.

weinbe58 (approved):

PR #606, opened 2026-05-04
Add "gemini_logical.star_rz(theta, reg)"

jasonhan3:
# Summary

Adds first-class Gemini Logical support for the Steane `[[7, 1, 3]]` STAR/TMR logical-Z injection rotation via `gemini.logical.star_rz(...)`.

The new operation is represented through the Gemini Logical, Place, and Move dialects, behaves like a logical single-qubit operation for placement, and lowers during transversal rewriting into physical `move.LocalRz` gates on a validated Steane logical-Z support.

# What Changed

- Added public API:
  - `gemini.logical.star_rz(theta, q, qubit_indices=None)`
  - `gemini.logical.star_rz(theta, qs, qubit_indices=None)`
- Added Steane STAR support validation with default support `(4, 5, 6)`.
- Added Gemini Logical `StarRz` statement and logical validation handling.
- Added Place dialect `StarRz`.
- Added Move dialect `StarRz`.
- Extended logical-to-place and place-to-move rewrites for STAR.
- Extended placement analysis so `place.StarRz` uses the same no-move single-qubit placement path as `R`/`Rz`.
- Extended static-placement merge, split, reorder, and fusion support for `StarRz`.
- Added `RewriteStarRz` before generic transversal location rewriting.
- Added AtomInterpreter/address validation handling for `move.StarRz`.
- Added `demo/star_demo.ipynb` showing:
  - logical `star_rz` kernel
  - logical Move output
  - physical Move output after STAR expansion
  - noiseless physical Squin output
  - `tsim.Circuit` diagram

# STAR Lowering Behavior

For v1, `star_rz` assumes:

- Steane `[[7, 1, 3]]`
- `k = 3`
- literal/compile-time `theta`
- no built-in postselection or syndrome extraction

The transversal rewrite computes:

```python
theta_star = -copysign(
    2 * atan(abs(tan(theta / 2)) ** (1 / 3)),
    theta,
)
```

Then it replaces logical `move.StarRz` with a physical `move.LocalRz` over:

```python
physical_site = logical_site * 7 + support_index
```

For the default support, logical site `0` lowers to physical sites `(4, 5, 6)`.

# Validation

Added/updated tests for:

- default support accepted
- all seven Steane supports accepted
- invalid supports rejected
- Gemini Logical `StarRz` lowering to Place `StarRz`
- Place `StarRz` lowering to Move `StarRz`
- placement analysis treating `StarRz` as a no-move single-qubit operation
- transversal rewrite lowering `move.StarRz` to physical `move.LocalRz`
- end-to-end logical pipeline behavior
- preservation of `qubit_indices` through placement merge/reorder/split rewrites

# Manual Verification

Ran:

```bash
uv run pytest python/tests/gemini/test_star_rz.py \
  python/tests/rewrite/test_circuit2place.py \
  python/tests/rewrite/test_place2move.py \
  python/tests/rewrite/test_transversal.py \
  python/tests/analysis/placement/test_place_dialect_lookahead.py \
  python/tests/pipeline/test_logical_pipeline.py -v
```

Result:

```text
64 passed
```

Also ran:

```bash
uv run ruff check python/bloqade/lanes/rewrite/circuit2place.py \
  python/bloqade/lanes/rewrite/reorder_static_placement.py \
  python/bloqade/lanes/rewrite/split_static_placement.py \
  python/tests/gemini/test_star_rz.py
```

Result:

```text
All checks passed!
```

And:

```bash
uv run pyright python/bloqade/lanes/rewrite/circuit2place.py \
  python/bloqade/lanes/rewrite/reorder_static_placement.py \
  python/bloqade/lanes/rewrite/split_static_placement.py \
  python/tests/gemini/test_star_rz.py
```

Result:

```text
0 errors, 0 warnings
```

Validated the notebook code path by extracting and running the code cells from `demo/star_demo.ipynb`. The compiled task produces:

- no remaining `move.StarRz`
- physical `move.LocalRz` on sites `(4, 5, 6)`
- a 7-qubit non-Clifford `tsim` circuit

# Notes

`star_rz` only implements the STAR rotation layer. Postselection and syndrome extraction remain explicit user-programmed steps.


python/bloqade/gemini/logical/dialects/operations/_interface.py
```diff
@@ -26,3 +26,20 @@ def terminal_measure(
             where each inner list corresponds to the measurements of the physical qubits that make up each logical qubit.
     """
     ...
+
+
+@lowering.wraps(StarRz)
+def star_rz(
+    theta: float,
+    qubits: Qubit | ilist.IList[Qubit, Len],
+    qubit_indices: tuple[int, int, int] | tuple[int, ...] | None = None,
```
weinbe58:
```suggestion
    qubit_indices: tuple[int, int, int] =(...),
```

For this feature we should only support the k=3 version + lets add a good default value here.

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -51,3 +84,35 @@ class TerminalLogicalMeasurement(ir.Statement):
...
+    logical qubit SSA value or an ``IList`` of logical qubits. The final
+    logical-to-physical lowering computes the physical STAR angle for k=3 and
+    emits physical Rz rotations on ``qubit_indices``.
+    """
+
+    traits = frozenset({lowering.FromPythonCall()})
+    rotation_angle: ir.SSAValue = info.argument(types.Float)
+    qubits: ir.SSAValue = info.argument(types.Any)
+    qubit_indices: tuple[int, int, int] = info.attribute(
+        default=DEFAULT_STEANE_STAR_SUPPORT
```
weinbe58, line 75: use that `DEFAULT_STEANE_STAR_SUPPORT` in the interface as well.

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -51,3 +84,35 @@ class TerminalLogicalMeasurement(ir.Statement):
...
+        qubits: ir.SSAValue,
+        *,
+        qubit_indices: tuple[int, int, int] | tuple[int, ...] | None = None,
+    ):
+        super().__init__(
+            args=(rotation_angle, qubits),
+            args_slice={"rotation_angle": 0, "qubits": 1},
+            result_types=(),
+        )
+        self.qubit_indices = validate_steane_star_support(qubit_indices)
```
weinbe58: please overload the `check` method of this statement and validate the attribute there.

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -51,3 +84,35 @@ class TerminalLogicalMeasurement(ir.Statement):
...
+
+    ``rotation_angle`` is the target logical angle. ``qubits`` may be either one
+    logical qubit SSA value or an ``IList`` of logical qubits. The final
+    logical-to-physical lowering computes the physical STAR angle for k=3 and
+    emits physical Rz rotations on ``qubit_indices``.
+    """
+
+    traits = frozenset({lowering.FromPythonCall()})
+    rotation_angle: ir.SSAValue = info.argument(types.Float)
+    qubits: ir.SSAValue = info.argument(types.Any)
```
weinbe58:
```suggestion
    qubits: ir.SSAValue = info.argument(ilist.IListType[QubitType, Any])
```
This needs to be a qubit array value

python/bloqade/gemini/logical/validation/clifford/impls.py
```diff
@@ -123,6 +124,32 @@ def controlled_gate(
...
+            interp.first_gates[address_value.data] = False
+        else:
+            interp.add_validation_error(
+                stmt,
+                ir.ValidationError(
+                    stmt,
+                    "Cannot determine qubit address location for star_rz.",
+                ),
+            )
+        return ()
```
weinbe58: why do we need this?

python/bloqade/lanes/dialects/move.py
```diff
@@ -124,6 +125,30 @@ class LocalRz(StatefulStatement):
...
+        qubit_indices: tuple[int, int, int] | tuple[int, ...] | None = None,
+    ):
+        ir.Statement.__init__(
+            self,
+            args=(current_state, rotation_angle),
+            args_slice={"current_state": 0, "rotation_angle": 1},
+            result_types=(StateType,),
+        )
+        self.location_addresses = location_addresses
+        self.qubit_indices = validate_steane_star_support(qubit_indices)
```
weinbe58: you should not need to implement an init here.

weinbe58: use the `check` method to run validation on the `qubit_indices`.

python/bloqade/lanes/dialects/place.py
```diff
@@ -134,6 +135,30 @@ class Rz(QuantumStmt):
...
+        qubit_indices: tuple[int, int, int] | tuple[int, ...] | None = None,
+    ):
+        ir.Statement.__init__(
+            self,
+            args=(state_before, rotation_angle),
+            args_slice={"state_before": 0, "rotation_angle": 1},
+            result_types=(StateType,),
+        )
+        self.qubits = qubits
+        self.qubit_indices = validate_steane_star_support(qubit_indices)
```
weinbe58: use the default initialization and use the `check` method to validate the statement.

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -1,10 +1,43 @@
...
+        raise ValueError("qubit_indices must contain integer physical qubit indices")
+    if out not in VALID_STEANE_STAR_SUPPORTS:
+        valid = ", ".join(
+            str(support) for support in sorted(VALID_STEANE_STAR_SUPPORTS)
+        )
+        raise ValueError(
+            f"qubit_indices must be a valid Steane weight-3 logical-Z support; "
+            f"got {out}. Valid Steane supports are: {valid}"
+        )
+    return cast(tuple[int, int, int], out)
```
weinbe58: This is duplicate of bloqade.lanes.star please remove the duplicate code.

python/bloqade/gemini/logical/dialects/operations/_interface.py
```diff
@@ -26,3 +26,20 @@ def terminal_measure(
             where each inner list corresponds to the measurements of the physical qubits that make up each logical qubit.
     """
     ...
+
+
+@lowering.wraps(StarRz)
+def star_rz(
+    theta: float,
+    qubits: Qubit | ilist.IList[Qubit, Len],
```
weinbe58:
```suggestion
    qubits: list.IList[Qubit, Len],
```

If you want a single argument version create a standard library function to simplify the rewrite logic

weinbe58 (changes requested):
Also make sure to look at the ASAP scheduler to make sure the reordering works for the new StarRz.

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -51,3 +84,35 @@ class TerminalLogicalMeasurement(ir.Statement):
...
+        rotation_angle: ir.SSAValue,
+        qubits: ir.SSAValue,
+        *,
+        qubit_indices: tuple[int, int, int] | tuple[int, ...] | None = None,
+    ):
+        super().__init__(
+            args=(rotation_angle, qubits),
+            args_slice={"rotation_angle": 0, "qubits": 1},
+            result_types=(),
+        )
```
weinbe58: Do not create an init here just use the auto generated init.

python/bloqade/lanes/rewrite/transversal.py
```diff
@@ -82,6 +86,80 @@ def rewrite_Statement(self, node: ir.Statement):
...
+    def _physical_support(
+        self,
+        logical_address: LocationAddress,
+        support: tuple[int, int, int],
+    ) -> tuple[LocationAddress, ...]:
+        if logical_address.site_id >= 2:
+            raise ValueError(
+                f"star_rz expected a logical site id 0 or 1, got "
+                f"{logical_address.site_id}"
+            )
```
weinbe58: There should not be exceptions in rewrites. also why is `site_id=0, 1` bad here?

python/bloqade/lanes/rewrite/transversal.py
```diff
@@ -82,6 +86,80 @@ def rewrite_Statement(self, node: ir.Statement):
         return rewrite_abc.RewriteResult(has_done_something=True)
 
 
+@dataclass
+class RewriteStarRz(rewrite_abc.RewriteRule):
+    """Lower logical STAR rotations to physical local Rz rotations.
+
+    v1 supports only literal target logical angles and the k=3 Steane STAR
+    protocol.
```
weinbe58: Why is this the case? it shuold not be difficult to support here its just a regular move gate.

python/bloqade/lanes/star.py
```diff
@@ -0,0 +1,38 @@
...
+        raise ValueError(
+            f"qubit_indices must be a valid Steane weight-3 logical-Z support; "
+            f"got {out}. Valid Steane supports are: {valid}"
+        )
+    return cast(tuple[int, int, int], out)
+
+
+def steane_star_theta(theta: float) -> float:
+    magnitude = 2 * math.atan(abs(math.tan(theta / 2)) ** (1 / 3))
+    return -math.copysign(magnitude, theta)
```
weinbe58: we should have this in the kirin dialect math: `kirin.dialects.math`

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -1,10 +1,77 @@
...
+        raise ValueError(
+            f"qubit_indices must be a valid Steane weight-3 logical-Z support; "
+            f"got {out}. Valid Steane supports are: {valid}"
+        )
+    return cast(tuple[int, int, int], out)
+
+
+@dataclass(frozen=True)
+class FromPythonStarRzCall(lowering.FromPythonCall["StarRz"]):
```
jasonhan3: This is because Kirin doesn't support default arguments in kernels (https://github.com/QuEraComputing/kirin/issues/440). So the question is, how do we let the user have a default value for the qubit_indices (the qubits on which to do the STAR rotation)? For now, we have this custom lowering function from Python to the IR that does both the default argument handling as well as the logic that turns a call with just one qubit and wraps it in a list.

python/bloqade/gemini/logical/dialects/operations/stmts.py
jasonhan3: So we run the validation both on the statement on the StarRz statement in the gemini.logical dialect (which lives in bloqade.gemini) as well as the StarRz statement in the move and place dialects (which live in bloqade.lanes). If we put this function only in one module (bloqade.gemini or bloqade.lanes), I think this will cause a circular import (for example, if we put the function in bloqade.lanes and imported it in bloqade.gemini, then gemini would import lanes, but lanes also imports gemini, which would cause the circular import). So for now I am duplicating the code; I think this can be solved if we put the code in a top level folder (like python/bloqade) but this would create a new "module" implicitly (so they would import from another module "bloqade.star"), but I don't think that's what we want

python/bloqade/gemini/logical/dialects/operations/stmts.py
weinbe58: This is not the same as a default value in a statement this isssue is about default arguments in the kernel function. please just set the `attribute(defualt=...)` and check to make sure it works. 

int he `interface.py` add the default value to the `@wraps` stub function which is not the same as a kernel function.

python/bloqade/gemini/logical/dialects/operations/stmts.py
weinbe58: Why don't you just put the validation in bloqade.gemini.star and then import that in bloqade.lanes then?

python/bloqade/gemini/logical/dialects/operations/stmts.py
jasonhan3: Got it; just removed the logic for the default arguments for qubit_indices from "FromPythonStarRzCall"; but still kept the custom lowering pass to allow support for the case where the user calls "star_rz" with just Qubit or ilist[Qubit].

I think the problem is:

If we use a kernel as a wrapper to then call the `@wraps` stub function, then the user can't supply default arguments, as Kirin kernels can't have default arguments.

But if we instead use default args on the `@wraps` stub function, then having default args for `qubit_indices` works, except now it's trickier to have the frontend for the user to supply either `Qubit` or `IList[Qubit]` as the logical qubits to operate on (the wrapping of a `Qubit` to `IList[Qubit]` is typically done in the kernel function, which we are not using).

Still have to work on a better solution, but for now, the custom lowering call from Python to the IR statement takes care of the wrapping of a `Qubit` to `IList[Qubit]`, but this means that something like

`gemini_logical.star_rz(theta, reg: IList[Qubit])`

works, which doesn't exactly match squin I believe (I think `IList[Qubit]` for squin gates is only supported with the explicit `broadcast` standard library kernel).

I think I will stick to this implementation for now and work on addressing the other comments first

python/bloqade/gemini/logical/dialects/operations/stmts.py
jasonhan3: Ah thank you! That appears to work; just made the change [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/bddc5a62b364c47e0aa03721952f51cacff051de)

python/bloqade/gemini/logical/validation/clifford/impls.py
jasonhan3: Yeah so this should be marking the qubits as "having had a gate applied to them" so that we do not allow noncliffords to be applied after the "star_rz" gate. So the following should err:

        @gemini.logical.kernel(aggressive_unroll=True)
        def invalid():
            q = squin.qalloc(1)
            gemini.logical.star_rz(0.125, q[0])
            squin.u3(0.1, 0.2, 0.3, q[0])
            
Made the change to reuse the existing "check_first_gate" in this commit: [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/cddcbf52a77210e391be50f2810ba7624c2411ba)

python/bloqade/gemini/logical/dialects/operations/_interface.py
jasonhan3: Added a default value for qubit_indices as (4, 5, 6) (commit: [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/c58d645e4a14161c9aff265a12ac2e5c240a8d78))

python/bloqade/gemini/logical/dialects/operations/stmts.py
weinbe58: I see, maybe we just have two statements then? the problem is that you do not know the type of the SSAValue when you're lowering it from python so the logic here doesn't work. Only after running type inference can you know the type. 

My suggestion is to just leave it as is or add two statements, one for a single qubit and one for a list of qubits. This doesn't need to be very user friendly anyways the minimal requirement is to to be able to demostrate the star. gagdet

python/bloqade/lanes/dialects/place.py
weinbe58, line 237: This is how you implement a static check on the statement, please use this in the StarRz statement and remove the __init__.py

python/bloqade/gemini/logical/dialects/operations/stmts.py
jasonhan3: Implemented in commit [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/63e992e9162c2c1d1f9bd2878e8337a0a9f26b27)

python/bloqade/gemini/logical/dialects/operations/stmts.py
jasonhan3: Implemented in commit [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/c58d645e4a14161c9aff265a12ac2e5c240a8d78)

python/bloqade/lanes/dialects/move.py
jasonhan3: Removed init in this commit: [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/d386cc33e2c0407df605b950f4242a7fa1e80075)

jasonhan3: Added check method in this commit: [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/d386cc33e2c0407df605b950f4242a7fa1e80075)

python/bloqade/lanes/dialects/place.py
jasonhan3: Done [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/16951a398a16636ca66835175c7ed408b71ac2e4)

python/bloqade/gemini/logical/dialects/operations/_interface.py
jasonhan3: Partially implemented [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/c58d645e4a14161c9aff265a12ac2e5c240a8d78) ; the logic doesn't work in terms of enforcing types for the "reg" argument; this may need to be changed to make it robust

python/bloqade/gemini/logical/dialects/operations/stmts.py
jasonhan3: Done [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/63e992e9162c2c1d1f9bd2878e8337a0a9f26b27)

python/bloqade/lanes/rewrite/transversal.py
jasonhan3: Removed the exception [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/c9b2b8f27d6b938d64175f2fb1b8a44900d48e7c). `site_id = 0, 1` as the only ones being valid is supposed to mimic "steane7_transversal_map", as we only have two logical sites.

python/bloqade/lanes/rewrite/transversal.py
jasonhan3: Added kirin math in the IR [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/a86aa0129345b293dde6d70ae8406ca5d583e951), so angles as SSA values (and thus passed as arguments to the kernel) should work OK

python/bloqade/lanes/star.py
jasonhan3: Added kirin math in the IR [here](https://github.com/QuEraComputing/bloqade-lanes/pull/606/changes/a86aa0129345b293dde6d70ae8406ca5d583e951)

python/bloqade/gemini/logical/dialects/operations/stmts.py
```diff
@@ -51,3 +56,35 @@ class TerminalLogicalMeasurement(ir.Statement):
...
+    )
+
+    def check(self) -> None:
+        if self.qubits.type.is_structurally_equal(
+            types.Bottom
+        ) or self.qubits.type.is_subseteq(QubitType):
+            raise exception.StaticCheckError(
+                "star_rz expects qubits to be an ilist.IList[Qubit]. "
+                "For one logical qubit, wrap it as ilist.IList([q]); for a "
+                "register, pass the register directly."
```
jasonhan3: remove this

python/bloqade/gemini/logical/stdlib/broadcast.py
jasonhan3, line 1: make kernel with only theta,qubits instead

python/bloqade/gemini/logical/stdlib/star.py
jasonhan3, line 1: remove this or create kernel with single qubit argument and angle

python/bloqade/gemini/logical/__init__.py
jasonhan3, line 1: fix up imports after stdlib refactor

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -216,10 +217,11 @@ def __init__(self, sites_per_word: int = 16):
         self.sites_per_word = sites_per_word
 
     def emit(self, mt: ir.Method, no_raise=True) -> ir.Method:
-        out = mt.similar(dialects=mt.dialects.add(dialect))
+        out = mt.similar(dialects=mt.dialects.add(dialect).add(kmath.dialect))
```
jasonhan3: kmath should already be in the dialect group

python/bloqade/lanes/rewrite/reorder_static_placement.py
```diff
@@ -132,7 +143,14 @@ def rewrite_Statement(self, node: ir.Statement) -> abc.RewriteResult:
...
         for stmt in new_stmts:
-            remapped = stmt.from_stmt(stmt, args=(curr_state, *stmt.args[1:]))
+            attributes: dict[str, ir.Attribute] = {}
+            if isinstance(stmt, place.StarRz):
+                attributes["qubit_indices"] = ir.PyAttr(stmt.qubit_indices)
+            remapped = stmt.from_stmt(
+                stmt,
+                args=(curr_state, *stmt.args[1:]),
+                attributes=attributes,
+            )
```
jasonhan3: I think you don't need to worry about all the attribute as long as the qubit_indices is not changing (which it is not here) the `from_stmt` should only need the attributes you are mutating

python/bloqade/lanes/rewrite/transversal.py
```diff
@@ -82,6 +91,59 @@ def rewrite_Statement(self, node: ir.Statement):
...
+    def _physical_support(
+        self, logical_address: LocationAddress, support: tuple[int, int, int]
+    ) -> tuple[LocationAddress, ...] | None:
+        iterator = self.transform_location(logical_address)
+        if iterator is None:
+            return None
+
+        physical_addresses = tuple(iterator)
+        if any(index >= len(physical_addresses) for index in support):
+            return None
```
jasonhan3: this should be caught by statement validation

python/bloqade/lanes/logical_mvp.py
```diff
@@ -128,6 +129,7 @@ def compile_squin_to_move(
         insert_return_moves=insert_return_moves,
     ).emit(mt, no_raise=no_raise)
     if transversal_rewrite:
+        mt = mt.similar(mt.dialects.add(kmath.dialect))
```
jasonhan3: double check again

python/bloqade/lanes/star.py
jasonhan3, line 1: dead code

python/tests/gemini/test_star_rz.py
```diff
@@ -0,0 +1,238 @@
...
+
+VALID_STEANE_STAR_SUPPORTS = {
+    (0, 1, 5),
+    (0, 2, 4),
+    (0, 3, 6),
+    (1, 2, 6),
+    (1, 3, 4),
+    (2, 3, 5),
+    (4, 5, 6),
+}
```
jasonhan3: use existing definition in bloqade.gemini

python/bloqade/lanes/rewrite/reorder_static_placement.py
jasonhan3: So for this case where qubit_indices is the only attribute that needs to be passed down, this can be removed; however, from_stmt does not seem to take in a "diff" for the "attributes" argument (which is what is being proposed) but rather takes in the new set of attributes; this means that if we change at least one attribute (ex: "qubits", which is changed in circuit2place.py and split_static_placement.py), then we need to supply a new attributes dict with all of the attributes that we want. (again, because from_stmt does not take in a 'diff' dict, but rather the new attributes dict to use)

python/bloqade/lanes/rewrite/split_static_placement.py
```diff
@@ -105,14 +105,19 @@ def rewrite_Statement(self, node: ir.Statement) -> abc.RewriteResult:
...
+                s
+                for s in block.stmts
+                if isinstance(s, (place.R, place.Rz, place.StarRz, place.CZ))
             ]
             for stmt in gate_stmts:
+                attributes: dict[str, ir.Attribute] = {
+                    "qubits": ir.PyAttr(tuple(remap[q] for q in stmt.qubits))
+                }
+                if isinstance(stmt, place.StarRz):
+                    attributes["qubit_indices"] = ir.PyAttr(stmt.qubit_indices)
```
weinbe58, line 117: do we need this still?

weinbe58 (approved):

PR #631, opened 2026-05-15
fix(search): gate entropy fallback on budget exhaustion

Jason-Ludmir:
## Summary
- Add entropy fallback tracing so `fallback_start` records the root node id, parent/depth, unresolved target count, root configuration, and best buffered resume node ids when sequential fallback begins.
- Surface fallback usage through `PhysicalPlacementStrategy.rust_entropy_fallback_count` by inspecting collected entropy trace events from solved Rust entropy stages.
- Change entropy search so sequential fallback only runs after the node expansion/iteration budget is exhausted and no goal node has been found, rather than when root entropy reaches `e_max`.

## Traceability
- Node identity: `node_id`, `parent_node_id`, and search `depth`.
- Search progress: `entropy`, `unresolved_count`, and `best_buffer_node_ids`.
- Search decisions: candidate movesets, candidate index, moveset score, and event reason fields already carried by trace steps.
- State diagnostics: repeated-state node ids, no-valid-move qubits, trigger node ids, current configuration, and parent configuration.
- Fallback usage: explicit `fallback_start` trace events and Python-level fallback counts for movement runs.

## Test Plan
- [x] `cargo test -p bloqade-lanes-search budget_exhaustion`
- [x] `cargo test -p bloqade-lanes-search root_entropy_limit_continues_without_sequential_fallback`
- [x] `PYTHONPYCACHEPREFIX=.pycache-check uv run --no-sync python -m py_compile python/bloqade/lanes/heuristics/physical/movement.py`
- [x] pre-commit hooks: end-of-files, trailing-whitespace, isort, black, ruff, pyright, cargo fmt, cargo clippy


weinbe58:
@Jason-Ludmir I assume the regression in the benchmarks is expected as now the entropy search is actually being bounded?

Jason-Ludmir:
> @Jason-Ludmir I assume the regression in the benchmarks is expected as now the entropy search is actually being bounded?

@weinbe58 Yes basically the regressions are mainly node increases, which makes sense:
Before, entropy would trigger fallback when root would hit e_max, meaning generally fewer nodes created.

After, fallback only triggers when node expansions hit the limit

weinbe58 (approved):
LGTM, just make sure to update the benchmark files.

PR #636, opened 2026-05-19
Gemini logical device API integration

david-pl:
@weinbe58 I'm not sure where I should put this stuff. Right now, I added a new module at `gemini/device`. I moved the `gemini/device.py` in that module and renamed it to `simulator.py`. I hope that doesn't make this a breaking change. CI still passes at least.


jasonhan3:
I think this looks good; I suppose in the future, we will have GeminiLogicalSimulator and GeminiLogicalSimulatorTask inherit from the generic Device and TaskABC?

david-pl:
> I suppose in the future, we will have GeminiLogicalSimulator and GeminiLogicalSimulatorTask inherit from the generic Device and TaskABC?

That's still TBD, I think, but at least for simulators that run behind the API that should be the case.

weinbe58 (approved):

PR #643, opened 2026-05-20
Feat: Adding option to remove LogicalTerminalMeasure when handling post processing removal. 

weinbe58:
In this PR I add a rewrite rule as well as an optional route to go through and remove any terminal measurement statement in the logical kernel. This is toggleable with a boolean flag and by default will do the original behavior making this change non-breaking.


python/bloqade/gemini/logical/rewrite/remove_postprocessing.py
```diff
@@ -82,7 +101,14 @@ class RemovePostProcessing(Pass):
     **NOTE**: Expects a flat logical kernel. Otherwise may lead to incorrect results.
     """
 
+    delete_terminal_measure: bool = field(kw_only=True, default=False)
+
     def unsafe_run(self, mt: Method) -> RewriteResult:
         result = Walk(_DeleteBelowTerminalMeasure()).rewrite(mt.code)
-        result = Walk(_InsertTerminalMeasureReturn()).rewrite(mt.code).join(result)
+        if self.delete_terminal_measure:
+            result = result.join(_DeleteTerminalMeasure().rewrite(mt.code))
```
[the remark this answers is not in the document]
weinbe58: Not needed this only triggers on when rewriting the Function Statement, no need to walk.

python/bloqade/gemini/logical/rewrite/remove_postprocessing.py
```diff
@@ -44,6 +44,24 @@ def rewrite_Statement(self, node: Statement) -> RewriteResult:
...
+    """Deletes the terminal measurement after the post processing
+    has been deleted. This MUST happen after _DeleteBelowTerminalMeasure
+    has been run on the kernel.
+    """
+
+    def rewrite_Statement(self, node: Statement) -> RewriteResult:
+        if not isinstance(node, func.Function):
+            return RewriteResult()
+
+        last_stmt = node.body.blocks[-1].last_stmt
```
david-pl, line 58: How is this not always a `func.Return` statement?

david-pl (approved):
LGTM

python/bloqade/gemini/logical/rewrite/remove_postprocessing.py
weinbe58: I run this after the `_DeleteBelowTerminalMeasure` rewrite which deletes the return in that rewrite the rule but only if the kernel has post processing if not the rule doesn't delete the return which is a little confusing but we can handle that edge case.

PR #645, opened 2026-05-20
Add Gemini MVP Demo Notebook

jasonhan3:
- Created a folder `demo/gemini_mvp` that has the demo notebook.
- Also added two files for local testing; `local_qlam_mock.py` for spinning up a mock local QLAM server; and `mock-qlam-config.json` for the config needed to test the code locally.


david-pl:
@jasonhan3 Demo looks good! A few things:

* Once we have bloqade-core v0.6.3 out, update the dependency to that and remove every mention of `SQLiteStorage` to simplify. You won't need to set the `storage` kwarg anymore, it will just default to `DictStorage`.
* I'd suggest you do break up the `run_pipeline` function. At least when running the first kernel, it would be nice to walk through each step, IMO. You can still keep the wrapper function for the following two kernels, but walking through everything once (and jupyter already gives you the ability to run each step separately), is kind of nice I think. It also gives you more "checkpoints" if something fails in between.
* Minor nit: instead of explicitly passing in `program_language: str`, you could just use the `kernel.sym_name`.

weinbe58 (approved):

PR #649, opened 2026-05-20
Fix: Adding test to check for ordering of final statements

weinbe58:
In pr #643 The insertion of the constant None value was after the terminal measurement which messed up the ordering of the SSA Values breaking the dominance principle. This PR adds a test to check for this as well as the fix to fix the insertion ordering.


david-pl (approved):
Also made sure the code downstream actually runs this time 😅 
Thanks for the quick fix!

johnzl-777 (approved):

PR #652, opened 2026-05-20
fix(visualizer): include blocked locations in entropy trace

Jason-Ludmir:
## Summary
- Minor update to the entropy tree visualizer trace data.
- Includes occupied non-layer qubits as blocked locations so traced layers render with the right context.

## Test plan
- uv run pytest python/tests/visualize/entropy_tree/test_tracer.py


weinbe58 (commented):
@Jason-Ludmir — the fix itself is correct and well-scoped. A few things to address:

---

## Blocking

### 1. Test should not depend on a file path — inline the kernel

```python
kernel_path = (
    Path(__file__).parents[3] / "benchmarks/kernels/large/trotter_rand_35.py"
)
kernel = load_kernel_from_file(kernel_path, "trotter_rand_35")
```

This couples the test to a file on disk. If `trotter_rand_35.py` moves, is renamed, or its contents change, the test breaks (or worse, silently passes with wrong counts). Please write a minimal inline kernel directly in the test — something like a 3-4 qubit circuit where you control exactly how many qubits are active vs. spectator in layer 0. That makes the test self-contained and the expected counts meaningful rather than magic numbers.

As a side effect, this also fixes the hardcoded `assert len(...) == 33` — the expected counts flow naturally from the inline circuit you define.

### 2. Mark `@pytest.mark.slow`

`build_entropy_trace` runs the full compilation pipeline and a Rust solve. Even with `max_expansions=5` it's in a different class from the unit tests above it. Per project convention, local runs use `pytest -m "not slow"` and the full suite only runs in CI.

---

## Should Fix

### 3. `default_factory=tuple` → `default=()`

```python
_traced_blocked_locations: tuple[LocationAddress, ...] = field(
    default_factory=tuple, init=False, repr=False
)
```

`()` is immutable so `default=()` is the correct form here. `default_factory=tuple` works but reads as if constructing from something.

### 4. Docstring: "non-local" → "spectator"

```python
"""Occupied non-local locations for the traced CZ layer."""
```

The codebase uses "spectator" for atoms not in the current CZ layer (see `EntanglingOptions.occupancy_penalty` docs, `entangling.rs`). Suggest: `"Spectator atom positions for the traced CZ layer (atoms not in the active placement)."`.

weinbe58 (changes requested):
@Jason-Ludmir see my comment on the PR directly.

Jason-Ludmir:
Fixed by replacing the file-backed fixture with a slow, self-contained inline spectator kernel test and updating the traced blocked-location field/default terminology to match spectator atoms.

weinbe58 (approved):
Thanks!

PR #663, opened 2026-05-26
fix(search): maximize AOD grouping across empty sites

### LINKED ISSUE #662: fix(search): maximize AOD rectangle grouping across empty sites
## Summary

AOD rectangle grouping is not selecting the maximal parallel move group in cases where the source/destination shape contains unoccupied holes. Instead of moving all compatible atoms in one AOD shot, the grouping can split the same logical rectangle into multiple smaller moves.

## Example

In the 17-qubit color-code custom layout, one observed CZ stage moves the leftmost logical block into the middle logical block. Each block spans four columns, with the fourth column only partially occupied. This should be schedulable as a single parallel AOD move because the occupied atoms in the group satisfy the parallelism constraints.

Observed behavior: the scheduler first selects the bottom two rows, then later selects the top three rows.

Expected behavior: the scheduler should select the full compatible group in one AOD rectangle/move, ignoring empty source/destination sites inside the rectangle when they do not correspond to moving atoms or collisions.

## Suspected Cause

The AOD grouping logic appears to require every coordinate in the candidate rectangle's Cartesian product to be present in the mover set. In `crates/bloqade-lanes-search/src/aod_grid.rs`, `BusGridContext::is_valid_rect` currently rejects a rectangle when any `(x, y)` position is absent from `movers`:

```rust
if !movers.contains(&src_enc) || self.collision_srcs.contains(&src_enc) {
    return false;
}
```

That seems too strict for partially occupied layouts. Empty source/destination sites inside an otherwise valid rectangle should not necessarily invalidate the AOD shot; only occupied/moving atoms and real collision risks should constrain the rectangle.

## Why This Matters

This causes non-maximal AOD grouping and introduces unnecessary sequential moves, especially for code layouts with partial columns/rows such as:

- Quantum Golay `[[23,1,7]]` layouts split across uneven rows
- 17-qubit color-code `[[17,1,5]]` layouts where the final column has only two qubits
- Similar custom physical layouts with sparse rectangular support

## Acceptance Criteria

- Add a regression case where a sparse but AOD-compatible rectangular move group is currently split into smaller row groups.
- Update AOD rectangle validation/grouping so empty source/destination positions inside a rectangle do not prevent grouping when they are not movers and do not collide.
- Preserve collision checks for occupied destinations and preserve correctness for fully occupied rectangles.
- Verify the affected debug layout produces a single maximal AOD group for the example stage.

Jason-Ludmir:
## Summary
- Allow sparse AOD rectangles to include empty filler lanes so lane groups remain complete while only selected mover atoms update the config.
- Reject unsafe filler lanes when the source has a spectator or the destination is occupied.
- Update receding-horizon replay to skip empty filler lane sources and add regression coverage for sparse AOD grouping.

Closes #662

## Test plan
- [x] `cargo fmt --all --check`
- [x] `cargo test -p bloqade-lanes-search aod_grid`
- [x] `cargo test -p bloqade-lanes-search receding_horizon`
- [x] `cargo test -p bloqade-lanes-search generate_candidates_emit_aod_rectangles`
- [x] `cargo test -p bloqade-lanes-search`
- [x] `uv run python debug/custom_layout/color_code_physical_51_entropy.py --no-interactive --max-expansions 20 --max-goal-candidates 1 --layer-index 0`


crates/bloqade-lanes-search/src/aod_grid.rs
```diff
@@ -259,6 +274,15 @@ mod tests {
...
         collisions: &[u64],              // src_encoded values with occupied destinations
+    ) -> BusGridContext {
+        make_context_with_occupied(positions, lanes, collisions, &[])
+    }
+
+    fn make_context_with_occupied(
+        positions: &[((u64, u64), u64)], // ((x, y), src_encoded)
+        lanes: &[(u64, u64)],            // (src_encoded, lane_encoded)
+        collisions: &[u64],              // src_encoded values with occupied destinations
+        occupied: &[u64],                // src_encoded values occupied by non-moving atoms
```
weinbe58: Just to be clear an unoccupied lane should have both the src and the dst not interact with stationary atoms. It is OK if the dst has a atom that is also moving within this bus.

weinbe58 (changes requested):

crates/bloqade-lanes-search/src/aod_grid.rs
Jason-Ludmir: Fixed: filler-lane validation now distinguishes stationary occupied destinations from destinations occupied by atoms moving in the same rectangle, and there is a regression test for that case.

weinbe58 (approved):
LGTM

PR #666, opened 2026-05-26
fix(benchmarks): exclude empty AOD lanes from move counts

Jason-Ludmir:
## Summary
- Count moved lanes from atom-state occupancy so empty AOD filler lanes are excluded from `move_count_lanes`.
- Keep move events unchanged and add a regression for empty filler lanes.
- Refresh `latest_physical.csv` with the locally regenerated physical benchmark baseline.

## Test plan
- `uv run pytest "python/tests/benchmarks/test_runner.py" -q`
- `uv run pyright "python/benchmarks/harness/runner.py" "python/tests/benchmarks/test_runner.py"`
- pre-commit hooks during commit


python/benchmarks/harness/runner.py
```diff
@@ -192,13 +200,23 @@ def _build_placement_strategy(self, job: BenchmarkJob):
...
+    frame, _ = atom_interp.run(move_mt)
     move_count_events = 0
     move_count_lanes = 0
     for stmt in move_mt.callable_region.walk():
         if isinstance(stmt, move.Move):
             move_count_events += 1
-            move_count_lanes += len(stmt.lanes)
+            state = frame.get(stmt.current_state)
+            if not isinstance(state, atom.AtomState):
+                continue
```
weinbe58: One question, do runs that fail to compile the program execute this code? If so I think this should be an assert because if we're skipping states that means there is a bug somewhere.

weinbe58 (commented):
Just left a comment no need to change anything at this point

python/benchmarks/harness/runner.py
Jason-Ludmir: Failed compiles return before this metric code runs, but I changed the non-AtomState case to an assert so analysis/state-threading bugs are surfaced instead of silently skipped.

weinbe58 (approved):

PR #668, opened 2026-05-27
Fix test for bloqade-core v0.6.3

david-pl:
Needs a release for https://github.com/QuEraComputing/bloqade-core/pull/73 first.

While the changes there are non-breaking, the test here is actually broken and needs to be fixed.


weinbe58 (approved):
LGTM Thanks!

PR #673, opened 2026-05-28
fix(gemini): adjust Steane transversal Clifford adjoints

rafaelha:
## Summary

For the Steane code, logical to physical translation is:
| Logical op            | Transversal physical op |
|-----------------------|-------------------------|
| `sqrt(X)`             | `sqrt(X)^dag`           |
| `sqrt(X)^dag`         | `sqrt(X)`               |
| `S`    | `S^dag`                 |
| `S^dag` | `S`                   |

Previously, the daggers were swapped, leading to a Pauli flip that was not tracked.

- Add a Steane-specific logical rewrite that flips the adjoint convention for transversal `sqrt_x` and `s` gates before logical native lowering.
- Wire the rewrite into the Gemini logical pipeline and cover both the standalone rewrite and pipeline integration with tests.

Note: The current treatment is specific to
- The Steane code
- And relies on the fact that non-Clifford gates like `rz, `u3` are not allowed (except for state initialization), even if their angle are half-pi multiples

## Why adjoint?
This convention is specific to the Steane code or any color code where logical ops have weight 3, 7, 11, ... (not color codes in general).

```text
X_L = X_1 X_2 X_3
Z_L = Z_1 Z_2 Z_3

the logical Pauli Y_L is

Y_L = i X_L Z_L
    = i (X_1 X_2 X_3)(Z_1 Z_2 Z_3)
    = i (X_1 Z_1)(X_2 Z_2)(X_3 Z_3)
    = i (-i Y_1)(-i Y_2)(-i Y_3)
    = i (-i)^3 Y_1 Y_2 Y_3
    = -Y_1 Y_2 Y_3.

Thus,

Y_1 Y_2 Y_3 = -Y_L.

Single-qubit Clifford conjugation signs therefore accumulate across the odd-weight transversal support. On one physical qubit,

sqrt(X) Z sqrt(X)^dag = Y.

So applying transversal physical sqrt(X) gives

Z_1 Z_2 Z_3 -> Y_1 Y_2 Y_3 = -Y_L,

which implements logical sqrt(X)^dag, up to the logical Pauli action convention. Applying transversal physical sqrt(X)^dag instead gives

Z_1 Z_2 Z_3 -> -Y_1 Y_2 Y_3 = Y_L,

which matches the desired logical sqrt(X) Pauli action.

The same odd-weight sign issue applies to sqrt(Z) and any Clifford gates that maps X/Z to Y.

### CODE REVIEW

File: python/bloqade/lanes/logical_mvp.py [hunk 1]
```diff
@@ -82,6 +85,10 @@ def transversal_rewrites(mt: ir.Method):
 
     """
 
+    # For [[7,1,3]] Steane code, we need to account that logical sqrt-X and sqrt-Z are
+    # implemented as transversal sqrt-X-adj and sqrt-Z-adj, respectively.
+    rewrite.Walk(RewriteSteaneTransversalCliffordAdjoints()).rewrite(mt.code)
+
```
weinbe58: This is not where this should go,it should go in the `LogicalPipeline` object

weinbe58 (changes requested):

python/bloqade/lanes/logical_mvp.py
rafaelha: I refactored it

python/bloqade/lanes/pipeline/logical.py
```diff
@@ -102,4 +136,7 @@ def emit(self, mt: Method, no_raise: bool = True) -> Method:
             insert_initialize=True,
         ).emit(out, no_raise=no_raise)
 
+        if self.transversal_rewrite:
+            out = transversal_rewrites(out)
```
weinbe58, line 140: Can you inline this function?

weinbe58 (changes requested):

weinbe58 (approved):
we can clean up things later

PR #681, opened 2026-05-29
Jasonh/677 starrz radians to turns

### LINKED ISSUE #677: "star_rz" gate doesn't turn radians to turns
"star_rz" gate should turn the radians that the user supplies as input into turns to be consistent with the rest of the angles in the compiler pipeline

jasonhan3:
Fixes #677 ; converts radians to turns for star_rz gate from user level


weinbe58 (approved):
LGTM

Issue #685, opened 2026-05-29
fix(search): entropy returns no candidates when AOD rectangles collapse


Jason-Ludmir:
## Summary
Entropy search can enter a deadlock state where individual unresolved qubits have seemingly improving bus moves, but `generate_candidates` returns zero movesets because all AOD rectangles are invalidated.

## Problem
- Candidate scoring selects promising single-qubit lane moves.
- AOD rectangle formation/validation then rejects every grouped rectangle (often due to occupied spectator destinations/sources).
- The search records `no-valid-moves` and entropy-bumps even though lane-level progress options exist.

This makes intended entropy routing fail for circuits that should be solvable by a small number of expansions.

## Expected behavior
When rectangle generation collapses to empty, entropy search should still produce a deterministic deadlock-breaking candidate layer to continue progress instead of returning no candidates.

## Reproduction
- Run:
  - `uv run python debug/run_steane_entropy_tree.py --layer-index 0 --max-expansions 1000 --max-goal-candidates 3`
- Observe nodes with no candidate movesets despite unresolved qubits having outgoing lanes.

## Acceptance criteria
- Deadlock condition is detectable in trace/debug output.
- Entropy search emits at least one valid deadlock-breaking candidate layer instead of returning empty candidates.
- Candidate breaker avoids large disruptive movesets (cap moved unresolved qubits to `ceil(unresolved/2)`).

PR #686, opened 2026-05-29
fix(search): add entropy deadlock-breaker candidate fallback

Jason-Ludmir:
## Summary
- add a Tier-2 deadlock-breaker fallback when entropy candidate rectangle construction returns empty
- cap fallback movers to at most `ceil(unresolved/2)` unresolved qubits to avoid disruptive global resets/cycles
- tag entropy trace descend steps with `reason="deadlock-breaker"` so the entropy tree visualizer shows when Tier-2 was used

Closes #685

## What This Changes
When `generate_candidates(...)` finds no valid AOD rectangle movesets for a node:
1. build deadlock-breaker candidates from unresolved-qubit lane scores
2. keep unresolved movers capped at `ceil(unresolved/2)`
3. build valid AOD rectangles from those entries
4. choose the best rectangle by:
   - moved unresolved qubits (maximize, within cap)
   - summed per-lane score (`total_score`)
   - lexicographic encoded-lane tie-break
5. descend with that candidate and annotate trace with `reason="deadlock-breaker"`

Important: breaker selection uses summed per-lane lane scores as tie-break within the capped set; it does not re-run the normal multi-rectangle ranking competition because the breaker emits a single fallback candidate.

## Small Example (2-Qubit Deadlock)
- unresolved qubits: `qA`, `qB`
- top normal moves collide/lock each other (no valid rectangle survives)
- cap is `ceil(2/2)=1`, so breaker can move at most one unresolved qubit
- breaker picks a legal side-step rectangle moving one qubit (e.g. `qB: B -> C`) based on ranking above
- search descends to the new config and can continue from a non-deadlocked state

## Test plan
- [x] `cargo test -p bloqade-lanes-search generate_candidates_ -- --nocapture`
- [x] `cargo test -p bloqade-lanes-search entropy_trace_marks_deadlock_breaker_descend -- --nocapture`

### DIFF
--- a/crates/bloqade-lanes-search/src/entropy.rs
+++ b/crates/bloqade-lanes-search/src/entropy.rs
@@ -117,7 +117,7 @@ impl Default for EntropyParams {
 struct EntropyState {
     entropy: u32,
     candidates_tried: usize,
-    candidate_cache: Vec<(MoveSet, Config, f64)>,
+    candidate_cache: Vec<CandidateEntry>,
     /// Encoded lane vecs of movesets already attempted from this node.
     tried_moves: HashSet<Vec<u64>>,
     /// Encoded lane vecs of movesets that failed (collision/transposition).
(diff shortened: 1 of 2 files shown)


weinbe58:
@Jason-Ludmir I am going to close this PR and incorporate it into my big refactor of the search infrastructure.

PR #691, opened 2026-05-30
fix(demo): make mock shot_index globally unique within a task

### LINKED ISSUE #690: Mock QLAM server: shot_index collision causes silent data loss for multi-subtask tasks
## Summary
`demo/gemini_mvp/local_qlam_mock.py` assigns each shot a `shot_index` local to its subtask (0..num_shots-1 per subtask). The SDK's SQLite storage enforces `UNIQUE(task_id, shot_index, frame_type)` with `INSERT OR IGNORE`. As a result, every subtask beyond index 0 in a `batch_task` or `parameter_scan` silently loses all its shots: rows collide with subtask 0's and are dropped without error.

## Reproduction

```python
from bloqade.core.device import SQLiteStorage
from bloqade.gemini import GeminiLogicalDevice
# kernels: bell (2q), ghz3 (3q), both terminal_measure

storage = SQLiteStorage(":memory:")
device = GeminiLogicalDevice(context_name="local-mock")
task = device.batch_task(kernels=[bell, ghz3], num_shots=4)
future = task.run_async(dry_run=False, storage=storage)
result = future.result(timeout=10)
# result.shot_results() -> [ndarray shape (4, 160), ndarray shape (0,)]
```

The mock's HTTP response carries all 8 shots; the loss is purely at the SQLite write boundary. Parameter scans exhibit the same loss for all subtasks > 0.

Inspecting the `results` table directly after `batch_task([bell, ghz3], num_shots=3)`:

```
task_id (short) | shot_idx | subtask_idx | subtask_shot_idx | frame_type
---------------------------------------------------------------------------
961e3294-4...   |    0     |      0      |        0         | DETECTED
961e3294-4...   |    1     |      0      |        1         | DETECTED
961e3294-4...   |    2     |      0      |        2         | DETECTED
```

Three rows total — subtask 1's three rows were dropped on `INSERT OR IGNORE`.

## Why this matters

- Silent: no error, no warning, no log line.
- Definition-level state is correct (the `subtasks` table has both rows with the right `program_index` and `num_shots`), so "did I submit it correctly?" checks pass.
- The notebook demo only uses single-kernel tasks, so the bug doesn't surface in casual exploration.

## Contract clarification

`python/tests/gemini/device/test_end_to_end.py::make_subtask` uses `shot_index = subtask_index * 1000 + i` (global within the task) and `subtask_shot_index = i` (local within the subtask). That is the intended contract — the SDK schema `UNIQUE(task_id, shot_index, frame_type)` is consistent with `shot_index` carrying the global identity. The production mock confuses `shot_index` with `subtask_shot_index`.

## Proposed fix

See linked PR. Compute a per-subtask cumulative offset and use `shot_offsets[subtask_index] + i` for `shot_index`.

Larger-surface alternatives worth considering separately:
- Add `subtask_index` to the SQLite `UNIQUE` key so future backends emitting local indices don't bite.
- Change `INSERT OR IGNORE` -> `ON CONFLICT ... DO ...` with a `logger.warning` so future shape-mismatches are visible rather than silent.

## Environment

bloqade-lanes @ origin/main (sha 1fd5073).

yzcj105:
## Summary
- Mock was assigning `shot_index = i` (local to subtask). SDK storage de-dupes by `UNIQUE(task_id, shot_index, frame_type)` with `INSERT OR IGNORE`, so all subtasks beyond index 0 silently lost their shots.
- Compute a per-subtask cumulative offset so `shot_index` is unique across the whole task, matching the contract used by the SDK test helper at `python/tests/gemini/device/test_end_to_end.py::make_subtask` (`shot_index = subtask_index * 1000 + i`).
- `subtask_shot_index` continues to be the per-subtask local index.

Fixes #690.

## Repro

Before this patch:

```python
task = device.batch_task(kernels=[bell, ghz3], num_shots=4)
future = task.run_async(dry_run=False, storage=storage)
result = future.result(timeout=10)
# result.shot_results() -> [ndarray (4, 160), ndarray (0,)]   # F14 data loss
```

After this patch:

```python
# result.shot_results() -> [ndarray (4, 160), ndarray (4, 160)]
```

## Test plan
- [ ] Repro the bug on `main`: `batch_task([bell, ghz3], num_shots=4)` returns one empty subtask in `result.shot_results()`.
- [ ] After this patch: both subtasks return `(4, 160)` arrays.
- [ ] Existing `python/tests/gemini/device/` suite still passes (those helpers already use globally-unique `shot_index`, so they are unaffected).


david-pl:
@yzcj105 did you verify that the demo notebook as it is right now on the main branch runs with this change?

weinbe58 (approved):
LGTM

yzcj105:
> @yzcj105 did you verify that the demo notebook as it is right now on the main branch runs with this change?

Verified on main branch commit 1fd5073

yzcj105:
@david-pl This is a fix for the mock test not for the frontend itself.

david-pl (approved):

weinbe58:
@david-pl I think this should be backported do you agree?

PR #697, opened 2026-06-02
feat: add Gemini decoding core package

jasonhan3:
## Dependencies

Same-repo stack:
- Code/API dependencies: none within `bloqade-lanes`; base is `jasonh/stdlib_firstit`.
- Merge order: first lanes PR in the stack.
- This PR is base for: #698.

Cross-repo dependencies:
- Direct code/API dependencies:
  - QuEraComputing/bloqade-decoders#36 for `bloqade.decoders.SparseTableDecoder`, imported by `bloqade.gemini.decoding.types.TableDecoderClass`.
  - QuEraComputing/bloqade-circuit#783 for `bloqade.analysis.tomography.DEFAULT_TARGET_BLOCH`, imported by `bloqade.gemini.decoding.constants`.
- Required for CI: yes, this PR imports the new decoder/tomography public APIs and declares `bloqade-decoders`.

Review status:
- Draft until dependencies are merged/released: yes.

## Summary

- Adds the `bloqade.gemini.decoding` package shell.
- Adds constants, shared types, layout helpers, and measurement-map helpers.
- Adds lazy public facades and declares `bloqade-decoders~=0.6.0`.

## Review focus

- Public module layout and dependency boundary.
- Lazy imports in package facades.

## Review Notes
- There is a monstrous amount of exports in the [__init__.py file](https://github.com/QuEraComputing/bloqade-lanes/pull/697/changes#diff-7074f2a014c04fa6289c872d10d37b97886efec016c7011227e272dc4ded8cdb) so that future PR's don't have to update this file (we use a dictionary of strings to support lazy imports... but this might not be the best design)
- `DetectorErrorModelTask` is used in `build_mle_decoders` to specify a task that just exposes a DEM to construct MLE decoders.

## Test plan

- Final lanes stack: `uv run --index-strategy=unsafe-best-match pytest python/tests/gemini/test_decoding_imports.py python/tests/demo/test_msd_utils.py -q` -> `41 passed in 18.46s`.
- Final lanes stack: targeted ruff command -> passed.
- Final lanes stack: `uv run --index-strategy=unsafe-best-match pyright python/bloqade/gemini/decoding` -> `0 errors, 0 warnings`.

PR #698, opened 2026-06-02
feat: add Gemini decoding task sampling

jasonhan3:
## Dependencies

Same-repo stack:
- Code/API dependencies:
  - #697 for the `bloqade.gemini.decoding` package shell and public facade this PR extends.
- Merge order: intended to merge after #697.
- This PR is base for: #699.

Cross-repo dependencies:
- Direct code/API dependencies introduced by this PR: none.
- Inherited from #697:
  - QuEraComputing/bloqade-decoders#36 for `SparseTableDecoder` in the table-decoder type facade.
  - QuEraComputing/bloqade-circuit#783 for `DEFAULT_TARGET_BLOCH`.
- Required for CI: yes, because the stacked branch includes #697's imports/dependency declaration.

Review status:
- Draft until dependencies are merged/released: yes.

## Summary

- Adds `GeminiDecoderTask`, `ObservableFrame`, `BasisDataset`, and chunked `run_task`.
- Adds CliffT/tsim sampling support.

## Review focus

- Observable frame normalization.
- Sampling chunking behavior and optional CliffT paths.

## Test plan

- `uv run --index-strategy=unsafe-best-match pytest python/tests/gemini/test_decoding_imports.py python/tests/demo/test_msd_utils.py -q` -> `41 passed in 18.46s`.
- Targeted ruff command -> passed.
- `uv run --index-strategy=unsafe-best-match pyright python/bloqade/gemini/decoding` -> `0 errors, 0 warnings`.

### DIFF
--- a/python/bloqade/gemini/decoding/sampling.py
+++ b/python/bloqade/gemini/decoding/sampling.py
@@ -1,0 +1,295 @@
+from __future__ import annotations
+
+from collections.abc import Iterator
+from dataclasses import dataclass
+from typing import Literal, Protocol, overload
+
+import numpy as np
+
+from .tasks import DemoTask, ObservableFrame
+
+
+@dataclass(frozen=True)
+class BasisDataset:
+    """Detector and observable samples for one tomography basis.
+
+    Attributes:
+        detectors: Detector samples with shape ``(shots, num_detectors)``.
+        observables: Observable samples with shape ``(shots, num_observables)``.
+    """
+
+    detectors: np.ndarray
+    observables: np.ndarray
+
+
+class DetectorObservableResult(Protocol):
+    @property
+    def detectors(self) -> object: ...
+
+    @property
+    def observables(self) -> object: ...
+
+
+class SimulatorTask(Protocol):
+    @overload
+    def run(
+        self,
+        shots: int,
+        with_noise: bool = True,
+        *,
+        run_detectors: Literal[False] = ...,
+    ) -> object: ...
+
+    @overload
+    def run(
+        self,
+        shots: int,
+        with_noise: bool = True,
+        *,
+        run_detectors: Literal[True],
+    ) -> DetectorObservableResult: ...
+
+    @overload
+    def run(
+        self,
+        shots: int,
+        with_noise: bool = True,
+        *,
+        run_detectors: bool,
+    ) -> object | DetectorObservableResult: ...
+
+
+# TODO: This is a wrapper to support both GeminiLogicalSimulatorTask and DemoTask.
+# Ideally, we can substitute this with something that better allows the user to
+# swap either using tsim or clifft as simulator backends.
+def _run_simulator_task(
+    task: SimulatorTask,
+    shots: int,
+    *,
+    with_noise: bool,
+    run_detectors: bool,
+    sim_type: str,
+) -> DetectorObservableResult:
+    """Run a simulator task through the selected backend."""
+
+    if not run_detectors:
+        raise ValueError("_run_simulator_task is only used for detector sampling.")
+    if isinstance(task, DemoTask):
+        return task.run(
+            shots,
+            with_noise=with_noise,
+            run_detectors=run_detectors,
+            sim_type=sim_type,
+        )
+    if sim_type != "tsim":
+        raise ValueError(
+            f"sim_type is {sim_type}; currently, the only supported simulator "
+            "backends are 'tsim' and 'clifft'"
+        )
+    return task.run(shots, with_noise=with_noise, run_detectors=run_detectors)
+
+
+# TODO: has separate clifft explicit check to avoid converting from
+# np.array -> list -> np.array. I think the problem is that DetectorResult
+# contains lists, so we have to do conversions to numpy arrays. So that's why
+# we are currently calling a clifft method that returns np.array's directly.
+# To change this, we'd need to change the task.run() interface/the DetectorResult
+# return type.
+def _sample_task_raw(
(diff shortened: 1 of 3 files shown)


python/bloqade/gemini/decoding/sampling.py
```diff
@@ -0,0 +1,295 @@
...
+from dataclasses import dataclass
+from typing import Literal, Protocol, overload
+
+import numpy as np
+
+from .tasks import DemoTask, ObservableFrame
+
+
+@dataclass(frozen=True)
+class BasisDataset:
```
jasonhan3, line 13: Is the datatype that we ingest for downstream analysis

python/bloqade/gemini/decoding/sampling.py
```diff
@@ -0,0 +1,295 @@
...
+    Attributes:
+        detectors: Detector samples with shape ``(shots, num_detectors)``.
+        observables: Observable samples with shape ``(shots, num_observables)``.
+    """
+
+    detectors: np.ndarray
+    observables: np.ndarray
+
+
+class DetectorObservableResult(Protocol):
```
jasonhan3, line 25: Is used to specify the interface of what we expect to be able to access from the result returned by the simulator (or other device)

python/bloqade/gemini/decoding/sampling.py
```diff
@@ -0,0 +1,295 @@
...
+    @property
+    def detectors(self) -> object: ...
+
+    @property
+    def observables(self) -> object: ...
+
+
+class SimulatorTask(Protocol):
```
jasonhan3, line 33: Used to define some interface for a task that can be run. I think bloqade-core exposes abstract Task interfaces, so maybe we can migrate to that instead.

python/bloqade/gemini/decoding/sampling.py
```diff
@@ -0,0 +1,295 @@
...
+        obs_chunks.append(np.asarray(result.observables, dtype=np.uint8))
+        remaining -= batch
+
+    return BasisDataset(
+        detectors=np.concatenate(det_chunks, axis=0),
+        observables=np.concatenate(obs_chunks, axis=0),
+    )
+
+
+def _compute_observable_reference(
```
jasonhan3, line 169: this is more of a helper function for us to get the observable reference in simulation and work off of that. In principle, this is not needed because we should know the observable reference when writing the circuit (for example, it should be all 0's).

python/bloqade/gemini/decoding/sampling.py
```diff
@@ -0,0 +1,295 @@
...
+) -> BasisDataset:
+    """XOR dataset observables by a reference observable row."""
+
+    return BasisDataset(
+        detectors=dataset.detectors,
+        observables=dataset.observables ^ reference.reshape(1, -1),
+    )
+
+
+def _normalize_observable_frame(
```
jasonhan3, line 209: In principle this is also not needed if you know that the observable reference is all 0's. But, because we were encountering "bugs" before where the observable wasn't all 0's, we have this automated fallback in case we don't know the observable reference

python/bloqade/gemini/decoding/tasks.py
```diff
@@ -0,0 +1,234 @@
...
+        observable_reference: Optional cached noiseless observable reference
+            used for rebasing special tasks.
+        metadata: Implementation metadata used by special task construction.
+    """
+
+    task: GeminiLogicalSimulatorTask[RetType]
+    observable_frame: ObservableFrame = ObservableFrame.RAW
+    observable_reference: np.ndarray | None = None
+    # TODO: this is SOLELY to pass down the "special" logical kernel for the "prefix_prepare" path.
+    metadata: dict[str, object] = field(default_factory=dict)
```
jasonhan3, line 57: In principle, we don't need `observable_frame`, `observable_reference`, and `metadata`.

`observable_frame` and `observable_reference` are used to specify if we need to do special handling to deal with nonzero observable references.

`metadata` is used to pass down a `special` logical kernel for the "prefix_prepare" path, but we can instead use the "compiled_inverse_prefix" path that doesn't need this `metadata`.

python/bloqade/gemini/decoding/tasks.py
```diff
@@ -0,0 +1,234 @@
...
+    ) -> Result[RetType] | DetectorResult:
+        sample_result = self._sample_clifft(shots, with_noise=with_noise, seed=seed)
+
+        fidelity_min, fidelity_max = self.task.fidelity_bounds()
+        if run_detectors:
+            return DetectorResult(
+                _detector_error_model=self.task.detector_error_model,
+                _fidelity_min=fidelity_min,
+                _fidelity_max=fidelity_max,
+                _detectors=np.asarray(sample_result.detectors, dtype=bool).tolist(),
```
jasonhan3, line 127: note that this is slow because we convert the numpy arrays back to lists. however, the current tsim simulator API returns lists I believe, so we want clifft to follow that same interface. we might want to change the API to return numpy arrays instead of lists.

python/bloqade/gemini/decoding/tasks.py
jasonhan3: Later on, if we get rid of `DemoTask` and just use `GeminiLogicalSimulatorTask`, both of these should expose a `run` method, so in principle it is interchangeable.

However, the current `clifft` integration is coupled with `DemoTask`, so we'd need to do some rethinking as to how to incorporate `clifft` into our pipeline.

jasonhan3:
After talking with Stefan, planning an alternative set of PR's to put up

PR #709, opened 2026-06-03
feat: updated gemini demo notebook to include first 7 columns only

jasonhan3:
- Created new notebook "gemini_mvp_demo_v2.ipynb" notebook with the following:
- Notebook now only subsets first 7 columns (as we are only filling half of the SLM grid)
- Deleted "plot_zero_probability_heatmap" function, as it was not used
- Addressed David's comments in this PR: https://github.com/QuEraComputing/bloqade-lanes/pull/645#issuecomment-4553179541
- Removed `SQLiteStorage` from the notebook (uses default `DictStorage`)
- Broke up the `run_pipeline` function for the pi/2 kernel
- Passed in `kernel.sym_name` instead of explicit `squin` program language


jasonhan3:
Addressed David's comments from this PR: https://github.com/QuEraComputing/bloqade-lanes/pull/645#issuecomment-4553179541

weinbe58 (approved):
Going forward can you use a jupytext fir developing the notebooks so it's a bit easier to review and get proper diffs?

PR #722, opened 2026-06-10
feat: made tsim lazy import for device/simulator.py

jasonhan3:
## Summary

Make `tsim` an optional dependency for `bloqade-lanes` while keeping Gemini logical device imports usable in environments where `tsim` is not installed.

- Move `bloqade-tsim` and the `bloqade-circuit[tsim]` extra out of the base dependency set and into a new `tsim` optional extra.
- Lazily import `tsim` only when a Gemini logical simulator task actually builds a tsim circuit.
- Preserve the `tsim_backend.Circuit` return annotations for simulator APIs using a `TYPE_CHECKING`-only import.
- Add regression coverage that imports `GeminiLogicalDevice` and `GeminiLogicalSimulator` without loading `tsim`.

## Details

`GeminiLogicalSimulator` still exposes `tsim_backend.Circuit` in type annotations, but the runtime dependency is now deferred behind `_tsim()`. If simulation code reaches the tsim circuit construction path without the optional extra installed, it raises an import error with an install hint:

```text
Install it with `bloqade-lanes[tsim]` or `uv sync --extra tsim`.
```

This keeps ordinary Gemini logical imports lightweight while preserving the existing simulator API shape for users who install the `tsim` extra.

## Testing

Ran:

```bash
uv run pytest python/tests/gemini/test_optional_tsim_import.py
uv run pyright python/bloqade/gemini/device/simulator.py python/tests/gemini/test_optional_tsim_import.py
uv run black --check python/bloqade/gemini/device/simulator.py python/tests/gemini/test_optional_tsim_import.py
```

Results:

```text
python/tests/gemini/test_optional_tsim_import.py: 2 passed
pyright: 0 errors, 0 warnings
black: files would be left unchanged
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,8 +11,7 @@ authors = [
     { name = "Phillip Weinberg", email = "pweinberg@quera.com" }
 ]
 dependencies = [
-    "bloqade-circuit[cirq, qasm2, tsim]~=0.14.1",
-    "bloqade-tsim~=0.1.3",
+    "bloqade-circuit[cirq, qasm2]~=0.14.1",
     "bloqade-core~=0.6.4",
     "deprecated>=1.3.1",
     "kirin-toolchain~=0.22.6",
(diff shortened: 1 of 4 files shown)


weinbe58 (approved):
LGTM

PR #725, opened 2026-06-11
feat: made cirq an optional dependency

jasonhan3:
## Summary

Make Cirq optional for the base `bloqade-lanes` install by moving Cirq-backed simulation/conversion dependencies into the `sim` extra.

- Remove the `cirq` extra from the required `bloqade-circuit` dependency.
- Add a `sim` optional extra that installs `bloqade-circuit[cirq, tsim]` and `bloqade-tsim`.
- Lazily import Cirq inside the QASM-to-SQUIN benchmark utility so importing the module does not require the optional `sim` extra.
- Update benchmark tests to use `pytest.importorskip("cirq")` only for tests that require Cirq.
- Refresh `uv.lock` to reflect the new optional-extra layout.

## Details

The QASM conversion utility still uses Cirq for parsing, native-gate conversion, qubit ordering, and statevector verification. Those paths now go through local imports, with a clearer install hint if the optional dependency is missing:

```text
Install it with `bloqade-lanes[sim]` or `uv sync --extra sim`.
```

Pure helpers in `test_qasm_to_squin_kernel.py` continue to run without Cirq installed, while Cirq-dependent tests are skipped in optional-free environments and run normally when CI/local installs include `--all-extras`.

## Testing

Ran:

```bash
uv run pyright python
uv run pytest python/tests/benchmarks/test_qasm_to_squin_kernel.py
uv run black --check python/benchmarks/utils/qasm_to_squin_kernel.py python/tests/benchmarks/test_qasm_to_squin_kernel.py
```

Results:

```text
pyright: 0 errors, 0 warnings
pytest: 2 passed, 3 skipped
black: 2 files would be left unchanged
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ authors = [
     { name = "Phillip Weinberg", email = "pweinberg@quera.com" }
 ]
 dependencies = [
-    "bloqade-circuit[cirq, qasm2]~=0.14.1",
+    "bloqade-circuit[qasm2]~=0.14.1",
     "bloqade-core~=0.6.4",
     "deprecated>=1.3.1",
     "kirin-toolchain~=0.22.6",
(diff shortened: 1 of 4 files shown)


weinbe58 (approved):
LGTM

PR #750, opened 2026-06-22
Add minimal experiments notebook

jasonhan3:
# Add Experiments Notebook
This PR creates a "msd_postselection_experiment.py" file (from which you can generate the notebook via jupytext).

The primary helper functions and API live in "bloqade/gemini/decoding/experiments.py". Other helper functions (a majority of which are private) live in "bloqade/gemini/decoding". 

## Change Summary
- Added "msd_postselection_experiment.py" python file
- Added bloqade/gemini/decoding module for helper functions supporting "msd_postselection_experiment.py"
- Modified "device/simulator.py" to add an "if" conditional as an ad-hoc implementation of configuring the simulation backend as CliffT instead of tsim. The idea is that the simulator object should encode what type of simulation backend we use. For the future, we want to address this in a more principled way.
- Export "decoding" module in bloqade/gemini/__init__.py
- Updated pyproject.toml to use bloqade-decoders>=0.6.0 which contains the MLE and MLD decoder implementations; thus, also bumped bloqade-circuit dependency to 0.14.3 to include the compatibility for MLE/MLD decoders. Also updated uv.lock as a result.

## Migration Notes and Breaking Changes
**Within-Module Migration Notes**
- The code in "tasks.py" should be moved to "bloqade/gemini/device/simulator.py" (or, a file in a similar location). It defines _CliffTSimulatorTask for CliffT integration and current inherits from GeminiLogicalSimulatorTask but defines additional fields for CliffT support and additional methods to obey the current GeminiLogicalSimulatorTask public API. However, some thought needs to be put into how to design the GeminiLogicalSimulatorTask to support different simulation backends (if this is a direction that we want to go in).

**Migration Notes**
- The code in "confidence.py" should be moved to bloqade-decoders. To prevent a breaking change, when the code is moved to bloqade-decoders, we can re-export the code in "confidence.py" and add a deprecation warning when a user tries to import from this path.
- The code in "dem.py" can be moved to bloqade-decoders (we can discuss if we would want to do this or not). To prevent a breaking change, we would do the same thing where we add a deprecation warning when the user tries to import from the current "dem.py" path.
- The code in "table_decoders.py" should be moved to bloqade-decoders and we will do the same thing to re-export the bloqade-decoders imports in "table_decoders.py".

**Expected Breaking Changes**
- In bloqade-decoders, we expect a breaking change in the __init__ function for TableDecoder. Currently, TableDecoder takes in "det_obs_counts: np.ndarray" as a required argument. For the public API for decoders, we likely just want to have Decoder(dem: stim.DetectorErrorModel); i.e., the stim.DetectorErrorModel should be the only required argument used to construct the decoder.
- In bloqade-decoders, we expect a breaking change in the default datatype used to store the counts. In particular, we may change the datatype from int64 to uint32 for space savings (uint32 still has a maximum value of around 4 billion, so this should be sufficient for a majority of use cases).
- In tandem with supporting different simulation backends (as was mentioned in "Within-Module Migration Notes"), we also might want to consider changing the Result return type in "simulator.py". The reason being, the Result type currently returns lists of raw measurements and lists of detectors/observables, but much of the code that I have for the PostSelectionExperiment operates on numpy arrays for faster vectorized operations. It may be beneficial to return these numpy arrays directly from the SimulatorTask API so that a user doesn't have to copy data from lists to numpy arrays which is a time overhead.


Krastanov (commented):
I am in favor of:

- the current "kitchensink" dump of functionality here, that is meant to be moved to other repos, done simply because there have been too many interdependencies to manage on a deadline
- the use of the experiment wizard style

I want to discuss:

- [x] tomography API
  - [x] filtering out low-confidence data
- [x] decoder API
  - [x] how initialization is done
  - [x] how confidence is reported
- [x] task launching API
- [x] analysis API
  - [x] and how it calls into the tomography API
- [x] preparation of useful subcircuits API
  - [x] how the various experiment/tomography objects manipulate circuits and compile them to tasks or DEMs
  - [x] how commonly used circuits are provided as a library

demo/msd_postselection_experiment.py
```diff
@@ -0,0 +1,198 @@
...
+# %% [markdown]
+# Here, we define three experiments: one for the lookup table decoder on the distillation circuit, one for a Gurobi Maximum-Likelihood Error decoder on the distillation circuit, and finally, one for the lookup table decoder on the experiment to obtain the injected fidelity of the encoded magic state.
+
+# %%
+msd_mld_exp = PostSelectionExperiment(
+    nonclifford_prefix,
+    clifford_circuit,
+    tomography_circuits,
+    TableDecoderWithConfidence,
+    {
```
Krastanov: nit: maybe name the argument so that the user knows what this optional thing is

demo/msd_postselection_experiment.py
```diff
@@ -0,0 +1,198 @@
...
+# %% [markdown]
+# ## Run the Experiments
+# On these "wizard" classes, we then call methods to initialize our decoders, obtain samples from our hardware device (for this demo, we are using a simulator in place of our hardware device), and performing decoding and postselection to analyze our results.
+
+# %%
+msd_mld_exp.kernels(num_logical_qubits=5)
+msd_mld_exp.dem_circuits()
+msd_mld_exp.dems()
+msd_mld_exp.initialize_decoders()
+msd_mld_exp.make_tasks(device=GeminiLogicalSimulator(backend="clifft", seed=10))
```
Krastanov, line 83: after rework should read 

`ClifftLogicalSimulator` or something like that, in order the pave the way for a uniform device/simulator interface

demo/msd_postselection_experiment.py
```diff
@@ -0,0 +1,198 @@
...
+msd_mld_exp.get_samples(num_shots=1_000_000)
+msd_mld_exp.decode_and_postselect(
+    np.array([[1, 0, 1, 1]]),
+    decoder_name="MLD",
```
Krastanov: nit: call it "progress_bar_prefix" or some similar but better short name

Krastanov: maybe "progress_bar_label"

Krastanov: True will lead to automatic choice based on str(type(the_decoder_instance))

demo/msd_postselection_experiment.py
```diff
@@ -0,0 +1,198 @@
...
+# > `accepted_fraction` is the fraction of shots you want to retain *after* the postselection criteria on the distillation circuit has been applied.
+#
+# > For example, say that you ran 1,000,000 shots for each basis. After decoding and postselection, we may only have 100,000 shots remaining in each basis. If you set `accepted_fraction` = 0.5, then you would accept around 150,000 shots across all three bases (note that the number of shots accepted in each basis may differ).
+
+# %%
+# The bloch vector to compute fidelity to (for MSD, we want to compute fidelity to the (1, 1, 1) state.)
+TARGET_BLOCH = np.ones(3, dtype=np.float64) / np.sqrt(3.0)
+
+tomo_result = msd_mld_exp.tomography_result(
+    0.50,
```
Krastanov: `accepted_fraction = 0.5`

Krastanov (approved):
Looks good, a couple of nits. Feel free to merge after the nits are addressed and tests pass.

I will make a few related issues for the planned rework and tag this PR.

Krastanov:
@weinbe58 , may I get write permissions.

I think we do not need an extra review -- things are well separated from the established codebase.

demo/msd_postselection_experiment.py
jasonhan3: ^ This will not be included in this PR and an issue should be tagged for this

python/bloqade/gemini/device/simulator.py
```diff
@@ -509,6 +513,21 @@ def task(
             post_processing,
         ) = compile_task(logical_kernel, m2dets, m2obs)
 
+        # TODO (Breaking): For the future, ideally we'd want to encode the simulator backend in the class itself or as a
+        # type parameter in the class. For now, we have a simple conditional to decide the simulator backend to use.
+        if self.backend == "clifft":
```
jasonhan3, line 518: @weinbe58 , this is the main change to the simulator.py code. It should be backwards compatible, as the default value of self.backend is "tsim".

python/bloqade/gemini/__init__.py
```diff
@@ -1,4 +1,4 @@
-from . import common as common, logical as logical
+from . import common as common, decoding as decoding, logical as logical
```
jasonhan3, line 1: @weinbe58 , exported "decoding" module from bloqade.gemini that has the code in this PR

PR #767, opened 2026-06-26
fix: made gurobipy, clifft, and tsim optional for pyright

jasonhan3:
- gurobipy, clifft, and tsim are optional for pyright, so making changes that allow pyright to deal with type checks for these optional imports


weinbe58 (approved):
LGTM

PR #786, opened 2026-07-07
fix(noise): honor GeminiNoiseModelABC scaling_factor in squin lowering

### LINKED ISSUE #785: Make sure upstream scale factor in GeminiNoiseModelABC is honored
Current behavior the function `generate_simple_noise_model` doesn't use the getter API's for getting the coefficients *with* the scale factors applied see the code reference: https://github.com/QuEraComputing/bloqade-circuit/blob/06116fd450aaeefebb4769bc42290f60ece02671/src/bloqade/cirq_utils/noise/model.py#L142

but is not possible to address the upstream issue: https://github.com/QuEraComputing/bloqade-circuit/issues/836 until the interface upstream is fixed. 

in the mean time we can fix the current issue in lanes and then enable a fix to get the correct noise model scaling upstream independently.

weinbe58:
## Summary

Fixes #785.

`generate_simple_noise_model` / `generate_logical_noise_model` read the **raw** `*_px` fields off the `GeminiNoiseModelABC` (e.g. `noise_model.mover_px`), which ignore the model's `scaling_factor`. The scaled values only live behind the getter properties (`mover_pauli_rates`, `two_qubit_pauli`, …). As a result, in the squin/stim path **every value of `scaling_factor` produced identical noise** — it worked at the cirq level but was silently a no-op downstream (see the discussion in QuEraComputing/bloqade-circuit#836).

## Changes

- Single-qubit Pauli rates now come from the scaled getters: `mover_pauli_rates`, `sitter_pauli_rates`, `local_pauli_rates`, `global_pauli_rates`, `cz_unpaired_pauli_rates`.
- Correlated CZ probabilities come from `two_qubit_pauli.error_probabilities` (scaled), with `.get(k, 0.0)` so a rate that scales to zero (and gets dropped from the dict) no longer risks a `KeyError`.
- **Loss probabilities are left raw on purpose** — upstream `scaling_factor` only scales Pauli rates, not loss. Per-category loss scaling is the genuinely-new axis tracked upstream in bloqade-circuit#836. Using the getters here also future-proofs lanes against that upstream work.
- New `python/tests/test_noise_scaling.py` walks the generated squin kernel IR and asserts the baked-in constants reflect `scaling_factor` across all noise kernels, that loss stays raw, that `scaling_factor=0` zeroes rates without `KeyError`, and (regression) that distinct scales now produce distinct constants.

## Verification

- `test_noise_scaling.py` + `test_noise_init_generator.py`: 37 passed
- `rewrite/move2squin/test_noise.py` + `test_device.py`: 27 passed, 3 skipped
- black / isort / ruff / pyright: clean


python/bloqade/lanes/noise_model.py
```diff
@@ -87,15 +88,18 @@ def idle_noise(qubits: ilist.IList[qubit.Qubit, Any]):
...
+    # scaled dict back off the channel. Missing keys (dropped when a scaled rate
+    # hits 0) default to 0.0 rather than raising.
+    cz_paired_error_dict = noise_model.two_qubit_pauli.error_probabilities
     cz_paired_error_probabilities = ilist.IList(
-        [cz_paired_error_dict[k] for k in PAIRED_KEYS]
+        [cz_paired_error_dict.get(k, 0.0) for k in PAIRED_KEYS]
     )
 
-    cz_unpaired_loss_prob = noise_model.cz_gate_loss_prob
+    cz_paired_loss_prob = noise_model.cz_gate_loss_prob
```
jasonhan3, line 102: shouldn't this be using cz_paired_errors on MoveNoiseModelABC if we want to use the properties instead of the direct field?

python/bloqade/lanes/noise_model.py
```diff
@@ -111,11 +115,9 @@ def pair_qubit(i: int):
 
         if loss:
             groups = ilist.map(pair_qubit, ilist.range(len(controls)))
-            squin.broadcast.correlated_qubit_loss(cz_unpaired_loss_prob, groups)
+            squin.broadcast.correlated_qubit_loss(cz_paired_loss_prob, groups)
```
jasonhan3, line 118: same comment; shouldn't this be using cz_paired_errors on MoveNoiseModelABC if we want to use the properties instead of the direct field?

jasonhan3 (changes requested):
In general, for `local_loss_prob, global_loss_prob, cz_gate_loss_prob, cz_unpaired_loss_prob, move_loss_prob, sit_loss_prob`, I don't know why you don't call the corresponding getters defined in `MoveNoiseModelABC` ? It's mentioned in the PR description that you don't but it's unclear to me why, as the issue here: https://github.com/QuEraComputing/bloqade-circuit/issues/836 has a `loss_error_scale` parameter.

python/bloqade/lanes/noise_model.py
weinbe58: That doesn't include the `scale` the API's are all messed up upstream

weinbe58:
Note on scope: this PR only scales the **Pauli rates**, not the **loss probabilities**. That's a deliberate limitation of the current upstream API — `GeminiNoiseModelABC.scaling_factor` (and its getter properties) only apply the scale factor to the Pauli rates; there is no getter that returns loss probabilities with the scale factor applied. So lanes has no way to obtain scaled loss values without re-implementing the scaling itself, which would drift from the upstream source of truth.

Per-category loss scaling (`loss_error_scale`) is the genuinely-new axis tracked upstream in QuEraComputing/bloqade-circuit#836. Once that lands and exposes a scaled-loss getter, we can switch the loss reads here over to it.

python/bloqade/lanes/noise_model.py
jasonhan3: Ah ok sounds good! So we'll probably update the downstream API's when the upstream ones are updated?

I guess, do we anticipate that the @property API's here: https://github.com/QuEraComputing/bloqade-circuit/blob/06116fd450aaeefebb4769bc42290f60ece02671/src/bloqade/qasm2/dialects/noise/model.py would have the scale factor? If so we can proactively change it; if not then we can just leave it alone until upstream API's are changed

jasonhan3 (approved):
LGTM!

PR #790, opened 2026-07-08
Refactor Gemini simulator tasks around backend objects

### LINKED ISSUE #789: refactoring Simulator and Task classes
This is another take on https://github.com/QuEraComputing/bloqade-lanes/pull/783

Currently  we have GeminiLogicalSimulator and PhysicalSimulator, both existing to mimic what the Gemini hardware would do in its two main modes, as a logical and as a physical quantum computer.

Both of these *Simulator classes take some configuration and then provide a `task` method. First refactor goal would be to move all configuration arguments from `task` to the constructor of the simulators, so that downstream consumers can be given a single *Simulator instance encapsulating the entirety of the configuration. `task` will only take the application circuit to be executed and no other arguments.

`task` will return a *Task instance. That instance will have a `run` method that returns samples. The `run` method should be consistent between task classes, taking same optional arguments, and only very few of them.

PhysicalSimulator should be renamed to GeminiPhysicalSimulator for consistency

Lastly, both *Simulator classes should be enabled to use different backend simulators, namely CliffT, Tsim, and PyQrack. The *Simulator class should take an instance of a subclass of AbstractSimulatorBackend. The *Task class will have a handled back onto the *Simulator class such that task.run() can pick a simulator backend. `AbstractSimulatorBackend` should have a simple API, as small as possible, to support its use from task.run().

The *Simulator object will continue being responsible for preparing the compilation pipeline and its `task()` method will be converting an "application level" squin circuit to a lanes / move dialect circuit. Then, given that this is a *Simulator class and not a "real device" class, that lanes / move dialect will be translated back to "noisy squin". Then `*Task.run()` will take the noisy squin and give it to the *SimulatorBackend to convert to whatever internal format it needs and to run samples on it.

Krastanov-agent:
Closes #789.

AI generated. Check my comments below. No self-contained prompt to share as there was some messy rework.


Krastanov-agent:
This PR deliberately keeps one uniform Gemini simulator result contract: Tsim, CliffT, and PyQrack tasks all provide a Tsim-derived detector error model. Consequently, CliffT and PyQrack currently compose `TsimSimulatorBackend`; the backend docstrings and dependency errors make that requirement explicit. A follow-up should separate native-program translation, sampling, and DEM generation into independent capabilities/providers so simulator backends do not need to depend on one another merely to satisfy the shared result contract. That broader separation is deferred here to keep this PR and each commit reviewable.

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -442,8 +442,14 @@ def task(
         from bloqade.lanes.passes import SequentialPlacePass
         from bloqade.lanes.pipeline import PhysicalPipeline
 
+        # Annotation, unrolling, and physical compilation all mutate their
+        # input. Keep the method supplied by the caller reusable.
+        source_squin_kernel = physical_kernel.similar()
```
Krastanov, line 447: This issue was detected by the AI -- repeated runs were modifying this repeatedly, making the kernel longer and longer.

Krastanov: the rest of the changes in this file are mostly just to address the 80 character line limit

Krastanov (commented):
- The first commit is purely for an issue detected by the AI where a kernel was repeatedly modified and made longer and longer each time the sim runs. I concur with the AI, but I do not have significant insight into it. Could you explicitly comment on the first commit?
- the second commit creates a base simulator backend and a tsim backend but does not use them for anything
- the third commit creates a clifft simulator backend BUT it requires a tsim backend because some of the necessary translation (not simulation) capabilities are inside of tsim (needs to be factored out in the future)
- the forth commit add pyqrack backend BUT it has the same need for tsim for non-simulation reasouns (needs to be factored out in the future)
- commits 2-4 can probably be treated more superficially in review
- commit five is the messy one where the previously defined backends are now used -- that one would need significant review though as I do not know whether overall this is done in an idiomatic manner; However, I am pretty happy that this commit deletes twice the number of lines it adds
- last commit -- move the configuration of the simulator away from `task` so that simulators can be consumed much more uniformly (task and run become uniform across all simulators) -- this would also require more focused review on whether it is idiomatic

One thing that I would like to request as a particular focus of the review is to let me know how you feel about the tests. (1) Some of the tests feel a bit redundant, as it is typical for AI-generated code. I am mostly ok with that, as I frequently read the tests first and seeing a deleted test in the future would make it easier for me to see minor changes; (2) generally, are they idiomatic for the way the team is already building the codebase -- any habits I am missing?

demo/cudaq_demo.py
```diff
@@ -15,7 +15,7 @@ def main_cuda():
     h(q[4])  # noqa: F821  # pyright: ignore
 
 
-task = GeminiLogicalSimulator().task(main_cuda, m2dets=m2dets, m2obs=m2obs)
+task = GeminiLogicalSimulator(m2dets=m2dets, m2obs=m2obs).task(main_cuda)
```
jasonhan3, line 18: Per a discussion today (and also after checking with Phil), generally for handling the m2dets/m2obs, I think we can have the user call "cudaq_to_squin" followed by "append_measurements_and_annotations" (which takes in the m2dets and m2obs) themselves. Thus task() should only support logical squin kernels, and `m2dets` and `m2obs` should not need to be passed to task() nor the GeminiLogicalSimulator() object at all.

jasonhan3:
> This PR deliberately keeps one uniform Gemini simulator result contract: Tsim, CliffT, and PyQrack tasks all provide a Tsim-derived detector error model. Consequently, CliffT and PyQrack currently compose `TsimSimulatorBackend`; the backend docstrings and dependency errors make that requirement explicit. A follow-up should separate native-program translation, sampling, and DEM generation into independent capabilities/providers so simulator backends do not need to depend on one another merely to satisfy the shared result contract. That broader separation is deferred here to keep this PR and each commit reviewable.

Yeah, so I also ran into this. The tsim repo defines some utilities for converting a circuit with nondeterministic observables to a detector error model (through converting the observables to detectors, then getting the DEM with nondeterministic detectors, then converting back to observables: https://github.com/QuEraComputing/tsim/blob/51d8350a5520107470c08cebf8219946de7c7591/src/tsim/noise/dem.py#L8). So what one could conceivably do is to basically just use stim with some helper functions for (a) the nondeterministic observables case and (b) supporting the nonclifford and atom loss/noise instructions. In this way, you can depend on stim + some additional utilities instead of depending on tsim.

However, these additional utilities are currently defined in the tsim repo. So all we'd need to do is move those utilities outside of tsim (maybe defining them in this repo or another dependency of bloqade-lanes) and import them then.

I've basically made this change to import the utilities from tsim for now in my refactoring PR: https://github.com/QuEraComputing/bloqade-lanes/pull/783, and in a future PR, we can move those utilities outside of tsim.

python/bloqade/gemini/device/physical_simulator.py
jasonhan3: So I think that this is only a problem if `m2dets` or `m2obs` is supplied and the user supplies a squin kernel (and the `m2dets` and `m2obs` will add annotations to the squin kernel). However, as per our discussion + checking with Phil, this annotation logic is only meant for CUDA-Q kernels, so we can move this annotation logic outside of "task". And if we move the `m2dets` and `m2obs` logic outside of the task() method, then we shouldn't need to copy the user's kernel here.

jasonhan3:
> The first commit is purely for an issue detected by the AI where a kernel was repeatedly modified and made longer and longer each time the sim runs. I concur with the AI, but I do not have significant insight into it. Could you explicitly comment on the first commit?

So I think if we make the design decision (based on our conversation today) that we have the user handle the CUDA-Q part on their own, then the changes in this first commit are not necessary, as I think it's only necessary to copy the kernel because we have this function that annotates the kernel with detectors/observables, which is only meant to be used for CUDA-Q supplied-kernels anyways.

python/tests/gemini/test_physical_simulator.py
```diff
@@ -123,15 +125,49 @@ def fake_append_measurements_and_annotations_physical(kernel, m2dets, m2obs):
...
+
+def test_physical_task_preserves_source_kernel_across_repeated_compilation():
+    @squin.kernel
+    def kernel():
+        reg = squin.qalloc(2)
+        return squin.broadcast.measure(reg)
+
+    source_ir = kernel.print_str()
+    simulator = PhysicalSimulator()
+    first = simulator.task(kernel, m2dets=[[1], [1]], m2obs=[[1], [0]])
```
jasonhan3, line 149: So the idea is that we would never call a squin kernel with m2dets/m2obs. The user should annotate the kernel themselves with squin.set_observable and squin.set_detector.

python/tests/gemini/test_physical_simulator.py
```diff
@@ -123,15 +125,49 @@ def fake_append_measurements_and_annotations_physical(kernel, m2dets, m2obs):
     )
 
     assert captured["placement_strategy"] is placement_strategy
-    assert captured["annotated_kernel"] is kernel
+    kernel.similar.assert_called_once_with()
```
jasonhan3, line 128: So these additions in the test file aren't needed because this feature isn't needed (see PR comment). This function just checks that similar() is called, which I guess is OK but is kind of awkward. I guess I'd just test directly that, after calling .task(), that my input kernel is the same structurally (by comparing the IR). I think this is what you do in the below test. (Although, this test itself is pretty awkwardly structured in the first place, so that might be the bigger problem)

python/tests/gemini/test_physical_simulator.py
```diff
@@ -123,15 +125,49 @@ def fake_append_measurements_and_annotations_physical(kernel, m2dets, m2obs):
...
+    assert captured["kernel"] is owned_kernel
     assert captured["no_raise"] is False
     assert captured["atom_dialects"] == "dialects"
     assert captured["post_processing_kernel"] is move_kernel
     assert task.physical_move_kernel is move_kernel
     assert task._post_processing == "post_processing"
+    assert task.source_squin_kernel is owned_kernel
+
+
+def test_physical_task_preserves_source_kernel_across_repeated_compilation():
```
jasonhan3, line 141: Again, this test is no longer needed as the associated feature of copying the kernel is no longer needed

python/tests/gemini/test_physical_simulator.py
```diff
@@ -123,15 +125,49 @@ def fake_append_measurements_and_annotations_physical(kernel, m2dets, m2obs):
...
+    def kernel():
+        reg = squin.qalloc(2)
+        return squin.broadcast.measure(reg)
+
+    source_ir = kernel.print_str()
+    simulator = PhysicalSimulator()
+    first = simulator.task(kernel, m2dets=[[1], [1]], m2obs=[[1], [0]])
+    second = simulator.task(kernel, m2dets=[[1], [1]], m2obs=[[1], [0]])
+
+    assert kernel.print_str() == source_ir
```
jasonhan3, line 152: So this is basically for the use case of "how do I test that one kernel equals another kernel?" One way is to print the string but I think this also gets to what you want to test (because it's possible for one kernel to be functionally equivalent to another one but have a different IR). However if you want to test that the entire IR and all the SSA are the same, then I think that kernel.print_str() is OK.

If you want to test that the kernels are "structurally" the same you can do `kernel.is_structurally_equal(kernel2)`.

I think you can also call `is_structurally_equal` on any `ir.Node` (so any Statement, Block, or Region in the IR).

For example, in the code, one pattern is that, to test if a rewrite works properly, you can do something like this

```
from bloqade.test_utils import assert_nodes
from kirin import ir, rewrite
from kirin.dialects import py
from kirin.rewrite import abc as rewrite_abc


class MyRewrite(rewrite_abc.RewriteRule):
    """Replace every py.Constant(1) with py.Constant(2)."""

    def rewrite_Statement(
        self,
        node: ir.Statement,
    ) -> rewrite_abc.RewriteResult:
        if isinstance(node, py.Constant) and node.value.unwrap() == 1:
            node.replace_by(py.Constant(2))
            return rewrite_abc.RewriteResult(has_done_something=True)

        return rewrite_abc.RewriteResult()


test_block = ir.Block([py.Constant(1)])

result = rewrite.Walk(MyRewrite()).rewrite(test_block)

expected_block = ir.Block([py.Constant(2)])

assert result.has_done_something
assert_nodes(test_block, expected_block)
```

I think the other tools for testing on Kirin kernels depend on your use case/exactly what it is that you're trying to test. Maybe @weinbe58 has more insight as to "idiomatic" ways to test functionalities on Kirin kernels

python/tests/gemini/test_physical_simulator.py
```diff
@@ -123,15 +125,49 @@ def fake_append_measurements_and_annotations_physical(kernel, m2dets, m2obs):
...
+            )
+            == 1
+        )
+        assert (
+            sum(
+                isinstance(stmt, SetObservable)
+                for stmt in task.source_squin_kernel.callable_region.walk()
+            )
+            == 1
+        )
```
jasonhan3, line 170: Checking the number of times a particular statement appears in the IR is another way to test

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -194,7 +181,7 @@ def _get_physical_measurement(q_idx: int, m_idx: int) -> ir.SSAValue:
 
 
 @dataclass(frozen=True)
-class PhysicalSimulatorTask(Generic[RetType]):
+class PhysicalSimulatorTask(_SimulatorTaskBase[RetType], Generic[RetType]):
```
jasonhan3, line 184: I think it might be possible to combine the physical/logical simulator classes and tasks into one, but we can do that in a future PR/discuss if this makes sense to do later.

jasonhan3:
A tracking comment for design decisions that impact the PR:
1. Do we want to combine the physical/logical compilation pipelines into one, and thus not distinguish between a physical and logical simulator? (My answer: it seems like this requires further discussion and we can maybe do this in a future PR)
2. Do we want to separate out the "run_detectors=True" functionality from the "def run" method defined on a task? Because this makes the return type be two different types depending on whether "run_detectors" is True or False.
- My answer: I think we can define an alternate method "def run_detectors" that explicitly samples the detectors and returns DetectorResult.
        - So if the user would instead want to run on the hardware and had code that used TsimSimulatorBackend, they would swap out the Simulator object with a Device object and call "run" instead of "run_detectors".
        - We would only define "run_detectors" on the TsimSimulatorBackend and on other backends that have a special sampling path for "run_detectors". For backends that don't have a separate "detector-sampling" mode (ex., CliffT), we wouldn't define "run_detectors".
3. Do we want to expose "tsim_circuit" on SimulatorTaskBase? This means that every simulator task is now dependent upon tsim.
       - A: We don't, and we should instead return a "stim_circuit", with helper functions as needed to convert the extra nonclifford and atom loss error channel annotations in the stim circuit to the format that we want.
4. What should handle the logic of converting from the squin kernel to a format that the backend can ingest?
       - A: Based on the current structure, I suppose that it makes sense for the backend object to do so (as the format that the backend ingests depends on the backend), but then I would say that we don't need to cache for now (as caching the conversion from squin -> backend input type is awkward)
                  - Technically, not caching might make existing user code slower if they have to recompile the circuit every time task() is called, but it's not "breaking" in terms of the API.
5. Do we want to support extracting detector error models from circuits with nondeterministic observables?
       - A: Maybe not; we can mirror stim's behavior. This would be a breaking change, though, because currently, we extract DEM's via tsim, which allows for circuits to have nondeterministic observables.
6. Does it make sense for PyQrackSimulator to emit a detector error model for arbitrary circuits where such circuits may have mid-circuit nonclifford gates?
       - A: not sure. How do errors propogate through nonclifford gates? Does it even make sense to generate a DEM from that?
7. Should the "seed" be a property on the simulation backend or every time we call sample()?
       - A: in a merged PR, we added the "seed" argument to run() and run_async(); we can probably do a "breaking" change to remove the "seed" argument from the CliffTSimulator (as it doesn't really make sense to have the argument on the simulator, and we had this before primarily just for quick integration)
8. Do we want some kind of "base" class/protocol to define shared implementations of task() wrappers from the logical/physical simulator classes?
       - A: ideally yes; saves us some redundant code

Auxiliary things/additional issues to make:
1. For PyQrackSimulatorTask and PyQrackSimulatorBase (and the simulator classes that inherit from it), it would be nice to integrate those simulator tasks and simulators together/put them together more cohesively. Maybe in the future we could have ONLY gemini-specific simulators in lanes and more generic simulators elsewhere.

Decisions we've made:
1. Move m2dets/m2obs to a separate method on the simulators to keep it separate from the "preferred" usage of the simulation.

Subtle notes:
1. For PPVM integration, which (I'm assuming) has loss-resolved readout, we will need to convert that atom loss to a boolean 0/1 for readout for gemini simulators.

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+class BackendSample:
+    """Raw samples returned by a simulator backend.
+
+    A backend may return measurements, or it may return detector and observable
+    arrays directly. Task execution validates and normalizes the selected form.
+    """
+
+    measurements: np.ndarray | None = None
+    detectors: np.ndarray | None = None
+    observables: np.ndarray | None = None
```
jasonhan3, line 26: To be clear: this is only the return type returned by the simulator backend (to standardize the type); however, in the current implementation, the return type returned by the *task* is different. I made a comment more regarding the latter in the design decisions tracking comment. Making a decision about the latter might help us to be able to more easily compose this BackendSample type returned by the simulatortask as well (as currently, I think that the return values from the *simulator backend* and the return values from the *simulator task* are different).

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+class AbstractSimulatorBackend(abc.ABC):
+    """Sampling and detector-model contract for Gemini simulators."""
+
+    @abc.abstractmethod
+    def sample(
+        self,
+        physical_squin_kernel: ir.Method,
+        *,
+        shots: int,
+        run_detectors: bool = False,
```
jasonhan3, line 38: I don't think this should be an option on the AbstractSimulatorBackend, as not all simulator backends really support run_detectors. `TsimSimulatorBackend` has different simulation options for whether `run_detectors` is True, but `CliffTSimulatorBackend` and `PyQrackSimulatorBackend` don't change the behavior in terms of simulation based on the value of `run_detectors`.

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+
+    @abc.abstractmethod
+    def detector_error_model(
+        self, physical_squin_kernel: ir.Method
+    ) -> DetectorErrorModel:
+        """Build the detector error model for a physical SQuIn kernel."""
+
+
+@runtime_checkable
+class _TsimCircuitCapability(Protocol):
```
jasonhan3, line 50: hmm.. is having @runtime_checkable desired? Ideally we would check this at compile time. I assume that we have this because we later want to remove this _tsim_circuit support and not depend on tsim explicitly but rather on stim potentially. This goes to point 3. on my design decisions comment.

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+    """Structural capability for backends that can provide a Tsim circuit."""
+
+    def _tsim_circuit(self, physical_squin_kernel: ir.Method) -> TsimCircuit: ...
+
+
+def _get_tsim_circuit(
```
jasonhan3, line 56: I assume that this method will be replaced once we decide on an interface for extracting a stim circuit + utility functions to analyze the stim circuit instead of returning tsim circuits?

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+        ) from exc
+
+    return tsim
+
+
+@dataclass
+class TsimSimulatorBackend(AbstractSimulatorBackend):
+    """Backend that uses Tsim for both sampling and DEM generation."""
+
+    _circuits: WeakKeyDictionary[ir.Method, TsimCircuit] = field(
```
jasonhan3, line 82: why do you have this?

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+
+        try:
+            return self._circuits[physical_squin_kernel]
+        except KeyError:
+            pass
+
+        owned_kernel = physical_squin_kernel.similar()
+        rewrite.Walk(RemoveReturn()).rewrite(owned_kernel.code)
+        circuit = _tsim().Circuit(owned_kernel)
+        self._circuits[physical_squin_kernel] = circuit
```
jasonhan3, line 98: why are we caching the tsim circuits generated from a squin kernel? this doesn't seem to be that useful... 

If @cached_property is defined on tsim_circuit on _SimulatorTaskBase, then why do we need to manually cache it in TsimSimulatorBackend ?

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+class _SimulatorTaskBase(Generic[RetType]):
+    """Shared execution runtime for compiled logical and physical tasks."""
+
+    physical_arch_spec: ArchSpec
+    physical_move_kernel: ir.Method[[], RetType]
+    _post_processing: atom.PostProcessing[RetType]
+
+    @property
+    def _backend(self) -> AbstractSimulatorBackend:
+        return cast(Any, self).simulator.backend
```
jasonhan3, line 105: How can you access a .simulator field on self if you don't guarantee that self will define a simulator field?

python/bloqade/gemini/device/simulator_backend.py
jasonhan3: I think that caching here is kind of awkward; maybe we can remove it for the first PR

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+@dataclass
+class TsimSimulatorBackend(AbstractSimulatorBackend):
+    """Backend that uses Tsim for both sampling and DEM generation."""
+
+    _circuits: WeakKeyDictionary[ir.Method, TsimCircuit] = field(
+        default_factory=WeakKeyDictionary, init=False, repr=False
+    )
+
+    def _tsim_circuit(self, physical_squin_kernel: ir.Method) -> TsimCircuit:
+        """Convert a physical SQuIn kernel to a cached Tsim circuit."""
```
jasonhan3, line 87: Note that implicitly here, we are saying that "the simulation backend is doing the conversion from squin to the input format to the simulator".

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+        circuit = _tsim().Circuit(owned_kernel)
+        self._circuits[physical_squin_kernel] = circuit
+        return circuit
+
+    def sample(
+        self,
+        physical_squin_kernel: ir.Method,
+        *,
+        shots: int,
+        run_detectors: bool = False,
```
jasonhan3, line 106: So tsim does have different behavior if you run the detector sampler or not, but other simulators do not. So I wouldn't make run_detectors part of the standard public API.

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -0,0 +1,146 @@
...
+
+        return BackendSample(
+            detectors=np.asarray(detectors, dtype=bool),
+            observables=np.asarray(observables, dtype=bool),
+        )
+
+    def detector_error_model(
+        self, physical_squin_kernel: ir.Method
+    ) -> DetectorErrorModel:
+        return self._tsim_circuit(physical_squin_kernel).detector_error_model(
```
jasonhan3, line 144: Hidden here is a custom functionality for extracting DEM's from a circuit with nondeterministic observables. We probably want to support circuits like these (?), OR we could introduce a breaking change and mirror stim's behavior where we raise an error if you try to get the detector error model on a circuit with nondeterministic observables. See design comment 5.

python/tests/gemini/test_simulator_backend.py
```diff
@@ -0,0 +1,163 @@
+from __future__ import annotations
```
jasonhan3, line 1: A lot of these tests are just testing that certain public API functions exist/are called, rather than testing for functional correctness, which I suppose is OK

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -144,3 +145,95 @@ def detector_error_model(
...
+
+    return clifft
+
+
+def _clifft_tsim_import_error(exc: ImportError) -> ImportError:
+    return ImportError(
+        "CliffT simulation also requires `bloqade-lanes[sim]`: Tsim performs "
+        "physical SQuIn conversion and detector error model generation even "
+        "when CliffT is selected for sampling."
+    )
```
jasonhan3, line 177: This relates to design point 3.; I suppose we have decided to refactor this in a future PR

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -144,3 +145,95 @@ def detector_error_model(
...
+
+
+@dataclass
+class CliffTSimulatorBackend(AbstractSimulatorBackend):
+    """Backend using Tsim for conversion/DEM generation and CliffT for sampling."""
+
+    seed: int | None = None
+    tsim_backend: TsimSimulatorBackend = field(
```
jasonhan3, line 185: maybe this should be private for now if we're going to deprecate this in the future?

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -144,3 +145,95 @@ def detector_error_model(
...
+        default_factory=TsimSimulatorBackend, repr=False
+    )
+    _programs: WeakKeyDictionary[ir.Method, Any] = field(
```
jasonhan3, line 188: Same comment on WeakKeyDictionary on TsimSimulatorBackend. Maybe the backend does not cache the tsim_circuit's generated from the squin kernels for now.

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -144,3 +145,95 @@ def detector_error_model(
...
+        default_factory=WeakKeyDictionary, init=False, repr=False
+    )
+
+    def _tsim_circuit(self, physical_squin_kernel: ir.Method) -> TsimCircuit:
```
jasonhan3, line 192: note this should be deleted in the future

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -144,3 +145,95 @@ def detector_error_model(
...
+        except ImportError as exc:
+            raise _clifft_tsim_import_error(exc) from exc
+
+    def _clifft_program(self, physical_squin_kernel: ir.Method) -> Any:
+        try:
+            return self._programs[physical_squin_kernel]
+        except KeyError:
+            pass
+
+        program = _clifft().compile(
```
jasonhan3, line 204: So ideally, as you mentioned in the PR comment, we would refactor out the dependency on _tsim_circuit and depend on stim only, with the extra nonclifford/noise instructions annotated

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -144,3 +145,95 @@ def detector_error_model(
...
+            self._clifft_program(physical_squin_kernel), **sample_kwargs
+        )
+        if run_detectors:
+            return BackendSample(
+                detectors=np.asarray(sample_result.detectors, dtype=bool),
+                observables=np.asarray(sample_result.observables, dtype=bool),
+            )
+        return BackendSample(
+            measurements=np.asarray(sample_result.measurements, dtype=bool)
+        )
```
jasonhan3, line 231: just a note that run_detectors is kind of pointless for the clifftsimulator because it doesn't change the simulation method, it just changes the return type

python/tests/gemini/test_simulator_backend.py
```diff
@@ -1,22 +1,28 @@
 from __future__ import annotations
 
 import gc
+import sys
```
jasonhan3, line 4: As this is just testing the API, I suppose this is reasonable

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -237,3 +238,123 @@ def detector_error_model(
...
+
+        if isinstance(node, (SetDetector, SetObservable)):
+            node.delete()
+            return RewriteResult(has_done_something=True)
+        return RewriteResult()
+
+
+@dataclass
+class PyQrackSimulatorBackend(AbstractSimulatorBackend):
+    """Backend using PyQrack for sampling and Tsim for guaranteed DEM generation."""
```
jasonhan3, line 266: Does it make sense for PyQrackSimulator (for arbitrary statevector/circuit simulation) to emit a detector error model with nonclifford operations in the middle of the circuit?

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -237,3 +238,123 @@ def detector_error_model(
...
+    _rng_state: np.random.Generator = field(init=False, repr=False)
+
+    def __post_init__(self) -> None:
+        self._rng_state = np.random.default_rng(self.seed)
+
+    def _tsim_circuit(self, physical_squin_kernel: ir.Method) -> TsimCircuit:
+        try:
+            return self.tsim_backend._tsim_circuit(physical_squin_kernel)
+        except ImportError as exc:
+            raise _pyqrack_tsim_import_error(exc) from exc
```
jasonhan3, line 283: again, hopefully we can refactor out the parts that convert squin to stim-text in a tsim-compatible format in a decoupled way where you don't have to import tsim.

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -237,3 +238,123 @@ def detector_error_model(
...
+        self,
+        physical_squin_kernel: ir.Method,
+        *,
+        shots: int,
+        run_detectors: bool = False,
+    ) -> BackendSample:
+        from bloqade.pyqrack.base import PyQrackInterpreter
+        from bloqade.pyqrack.device import StackMemorySimulator
+        from bloqade.pyqrack.reg import MeasurementResultValue
+        from bloqade.pyqrack.task import PyQrackSimulatorTask
```
jasonhan3, line 295: I didn't realize that PyQrack already had a simulatortask class.. maybe in a future PR we could more cohesively integrate/merge these "backends" with more generic simulators (that just ingest a squin kernel, agnostic to the compilation pipeline applied).

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -237,3 +238,123 @@ def detector_error_model(
...
+        owned_kernel = physical_squin_kernel.similar()
+        rewrite.Walk(_RemovePyQrackAnnotations()).rewrite(owned_kernel.code)
+        simulator = _RecordingStackMemorySimulator(
+            options=cast(Any, self.options or {}),
+            rng_state=self._rng_state,
+            min_qubits=self.min_qubits,
+        )
+        recorded_measurements: list[list[bool]] = []
+        for _ in range(shots):
+            task = simulator.task(owned_kernel)
```
jasonhan3, line 334: Why do you have to create a new task per shot? It might be better to move the task outside of the loop

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -237,3 +238,123 @@ def detector_error_model(
...
+            task.run()
+            interpreter = cast(_RecordingPyQrackInterpreter, task.pyqrack_interp)
```
jasonhan3, line 336: why do you have to cast here? oh, is it because you don't want to override the type of the field of task.pyqrack_interp? I suppose that's alright

demo/cudaq_demo.py
jasonhan3, line 18: instead of putting m2dets and m2obs as arguments to the constructor, we can define an additional method that takes these as inputs.

demo/msd_postselection_experiment.py
```diff
@@ -75,12 +75,15 @@
 # ## Run the Experiments
 # On these "wizard" classes, we then call methods to initialize our decoders, obtain samples from our hardware device (for this demo, we are using a simulator in place of our hardware device), and performing decoding and postselection to analyze our results.
 
+# %%
+clifft_simulator = GeminiLogicalSimulator(backend=CliffTSimulatorBackend(seed=10))
```
jasonhan3, line 79: we can get rid of support for the "seed" argument in the constructor of CliffTSimulatorBackend as in design decision 7. ; we can just support it in the run() and run_async() methods.

python/bloqade/gemini/decoding/experiments.py
```diff
@@ -222,8 +223,12 @@ def dem_circuits(self) -> dict[str, tsim_backend.Circuit]:
         dem_kernels = self._postselection_exp_cache.dem_kernels
         if dem_kernels is None:
             raise RuntimeError("kernels must be called before dem_circuits.")
+        dem_simulator = GeminiLogicalSimulator(
+            noise_model=self._simulator.noise_model,
+            backend=TsimSimulatorBackend(),
+        )
```
jasonhan3, line 229: why were these lines added? _simulator can already compile the task; there's no need to construct another one?

jasonhan3: I suppose that it's to explicitly use tsim? But I guess it's not strictly needed

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+
+    @property
+    def _physical_kernel(self) -> ir.Method[[], RetType]:
+        return cast(Any, self).physical_squin_kernel
+
+    @property
+    def _noiseless_physical_kernel(self) -> ir.Method[[], RetType]:
+        return cast(Any, self).noiseless_physical_squin_kernel
```
jasonhan3, line 113: why don't we instead just add these properties as fields in the class definition rather than making these private wrapper functions with casts that assume "simulator", "physical_squin_kernel", and "noiseless_physical_squin_kernel" exist?

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+
+    @cached_property
+    def tsim_circuit(self) -> tsim_backend.Circuit:
+        """Tsim-compatible circuit for the noisy physical kernel."""
+        return _get_tsim_circuit(self._backend, self._physical_kernel)
+
+    @cached_property
+    def noiseless_tsim_circuit(self) -> tsim_backend.Circuit:
+        """Tsim-compatible circuit for the noiseless physical kernel."""
+        return _get_tsim_circuit(self._backend, self._noiseless_physical_kernel)
```
jasonhan3, line 127: I wonder if every simulator task should expose a tsim circuit? Maybe we don't and instead want to just expose stim_circuit (see design decision 3.)

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+    def noiseless_measurement_sampler(self):
+        return self.noiseless_tsim_circuit.compile_sampler()
+
+    @cached_property
+    def detector_sampler(self):
+        return self.tsim_circuit.compile_detector_sampler()
+
+    @cached_property
+    def noiseless_detector_sampler(self):
+        return self.noiseless_tsim_circuit.compile_detector_sampler()
```
jasonhan3, line 143: Also related to design decision 3.: not sure if we want to expose these tsim-specific samplers on the base simulator task class.

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+
+    @cached_property
+    def detector_error_model(self) -> DetectorErrorModel:
+        """Guaranteed detector error model for the noisy physical kernel."""
+        return self._backend.detector_error_model(self._physical_kernel)
```
jasonhan3, line 148: So I think this is especially confusing because it looks like both the tasks and the backend can do the job of converting a squin kernel to a tsim circuit. We probably only want this to be done in ONE location and that location should probably be the backend, as each simulation backend could decide how to convert from squin to its input format (see design decision 4.)

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+    def run(
+        self,
+        shots: int = 1,
+        with_noise: bool = True,
+        *,
+        run_detectors: bool = False,
+    ) -> Result[RetType] | DetectorResult:
+        """Sample through the configured backend and normalize its result."""
+        # Build the guaranteed DEM before beginning a potentially expensive
+        # sampling request. This also fails early when Tsim is unavailable.
```
jasonhan3, line 227: as per design point 3., failing early when tsim is unavailable is probably not the desired behavior.

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+        detector_error_model = self.detector_error_model
+        physical_kernel = (
+            self._physical_kernel if with_noise else self._noiseless_physical_kernel
+        )
```
jasonhan3, line 231: just a note: is kind of confusing to read because you basically enforce that the subclasses implement physical_squin_kernel and noiseless_physical_squin_kernel. A better design might be an abstract property in the base class for physical_squin_kernel and noiseless_physical_squin_kernel and a cached property in the base classes for the concrete implementation of the properties?

python/bloqade/gemini/device/_task_runtime.py
```diff
@@ -0,0 +1,344 @@
...
+            )
+            observables = self._normalize_matrix(
+                sample.observables, name="observable", shots=shots
+            )
+        elif sample.measurements is not None:
+            measurements = self._normalize_matrix(
+                sample.measurements, name="measurement", shots=shots
+            )
+            detectors = list(self._post_processing.emit_detectors(measurements))
+            observables = list(self._post_processing.emit_observables(measurements))
```
jasonhan3, line 286: a bit of redundant logic to apply post_processing here as well as in the Result object type in order to convert the raw measurements to detectors/observables. see design decision 2.

python/bloqade/gemini/device/simulator.py
```diff
@@ -27,176 +31,28 @@
     from bloqade.lanes.rewrite.move2squin.noise import LogicalNoiseModelABC
 
 RetType = TypeVar("RetType")
+LogicalKernel = Union[ir.Method[[], RetType], Callable[..., Any]]
```
jasonhan3, line 34: note that this is only to support cuda-q; see "decisions we've made" 1. to explicitly separate out the cuda-q funcitonality

python/bloqade/gemini/device/simulator.py
```diff
@@ -27,176 +31,28 @@
...
     physical_arch_spec: ArchSpec = field(repr=False)
-    """The physical architecture specification."""
     physical_move_kernel: ir.Method[[], RetType] = field(repr=False)
-    """The physical move kernel that executes the logical squin kernel on the physical architecture."""
     _post_processing: atom.PostProcessing[RetType] = field(repr=False)
-    """The post-processing object for extracting detectors, observables, and return values."""
-    _thread_pool_executor: ThreadPoolExecutor = field(
-        default_factory=ThreadPoolExecutor, init=False
-    )
+    simulator: GeminiLogicalSimulator = field(repr=False)
```
jasonhan3, line 52: I wonder if you need the whole simulator in the GeminiLogicalSimulatorTask, or if passing down the simulator._backend is sufficient?

python/bloqade/gemini/device/simulator.py
```diff
@@ -216,330 +71,42 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
...
 
     noise_model: LogicalNoiseModelABC = field(default_factory=_default_noise_model)
-    """The noise model used for simulation. Defaults to :func:`generate_logical_noise_model`."""
-    backend: str = "tsim"
-    """Sampling backend for tasks created by this simulator."""
-    seed: int | None = None
-    """Optional backend seed for task sampling."""
+    backend: AbstractSimulatorBackend = field(default_factory=TsimSimulatorBackend)
+    m2dets: list[list[int]] | None = None
+    m2obs: list[list[int]] | None = None
```
jasonhan3, line 82: as per "decisions we've made" 1., we probably don't want to pass in m2dets and m2obs in geminilogicalsimulator.

python/bloqade/gemini/device/simulator.py
```diff
@@ -216,330 +71,42 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
...
-                squin or CUDA-Q kernel to compile and run.
-            m2dets (list[list[int]] | None): Binary measurement-to-detector matrix.
-                For CUDA-Q kernels, defaults to Steane [[7,1,3]] detectors if ``None``.
-            m2obs (list[list[int]] | None): Binary measurement-to-observable matrix.
-                For CUDA-Q kernels, defaults to Steane [[7,1,3]] observables if ``None``.
-
-        Returns:
-            GeminiLogicalSimulatorTask[RetType]: The compiled simulation task.
-
-        """
```
jasonhan3, line 506: we probably don't want to remove the docstrings

python/bloqade/gemini/device/simulator.py
```diff
@@ -216,330 +71,42 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
...
-        if self.backend != "tsim":
-            raise ValueError("backend must be either 'tsim' or 'clifft'.")
+        ) = compile_task(logical_kernel, self.m2dets, self.m2obs)
         return GeminiLogicalSimulatorTask(
             logical_squin_kernel,
             self.noise_model,
             physical_arch_spec,
             physical_move_kernel,
             post_processing,
+            self,
```
jasonhan3, line 103: instead of passing "self" here, maybe just pass in simulation backend so it's more explicit what is needed by the task.

python/bloqade/gemini/device/simulator.py
```diff
@@ -568,36 +135,20 @@ def run(
...
-            run_detectors (bool): When ``True``, use the detector sampler instead of
-                the measurement sampler for faster detector/observable sampling.
-                Defaults to False.
-
-        Returns:
-            Result[RetType]: When ``run_detectors=False``, the full simulation result.
-            DetectorResult: When ``run_detectors=True``, the result containing only
-                detector and observable outcomes.
-
-        """
```
jasonhan3, line 592: again, we don't want to delete our docstrings

python/bloqade/gemini/device/simulator.py
```diff
@@ -607,116 +158,63 @@ def run_async(
     @overload
     def run_async(
         self,
-        logical_squin_kernel: ir.Method[[], RetType],
+        logical_kernel: LogicalKernel[RetType],
```
jasonhan3, line 161: all of these changes to LogicalKernel should not need to happen as the changes to LogicalKernel are just for CUDA-Q support. see "Decisions we've made" 1.

python/bloqade/gemini/device/simulator.py
```diff
@@ -607,116 +158,63 @@ def run_async(
...
 
+    @overload
     def run_async(
         self,
-        logical_squin_kernel: ir.Method[[], RetType],
+        logical_kernel: LogicalKernel[RetType],
+        shots: int = 1,
+        with_noise: bool = True,
+        *,
+        run_detectors: bool,
```
jasonhan3, line 175: what's the point of this overload to not specify the value of run_detectors? I thought the whole point of the overloads here was to specify the return type from run_async

python/tests/gemini/test_optional_tsim_import.py
```diff
@@ -2,57 +2,61 @@
...
+            sys.executable,
+            "-c",
+            textwrap.dedent(code) + textwrap.dedent(OPTIONAL_BACKEND_IMPORT_CHECK),
+        ],
+        check=True,
+    )
+
+
+def test_logical_device_import_does_not_import_optional_backends():
+    _run_import_guard("""
```
jasonhan3, line 37: I probably already left a comment on the pyqrack/clifft simulator backends, but having them composed of tsim simulator backends is fairly unintuitive. See design decision 3. for a proposed fix to couple to stim instead of tsim.

python/tests/gemini/test_optional_tsim_import.py
```diff
@@ -2,57 +2,61 @@
...
-            raise SystemExit(f"unexpected tsim imports: {loaded_tsim}")
         """)
 
-    subprocess.run([sys.executable, "-c", code], check=True)
-
 
-def test_logical_simulator_import_does_not_import_tsim():
-    code = textwrap.dedent("""
+def test_logical_simulator_import_does_not_import_optional_backends():
+    _run_import_guard("""
```
jasonhan3, line 49: this just checks for lazy imports; for eager imports I think all backends will require importing tsim which might not be ideal

python/tests/gemini/test_optional_tsim_import.py
```diff
@@ -65,16 +69,58 @@ def test_tsim_backend_import_is_lazy():
...
-        assert not any(
-            name == "tsim"
-            or name.startswith("tsim.")
-            or name == "bloqade.tsim"
-            or name.startswith("bloqade.tsim.")
-            for name in sys.modules
+        """)
+
+
+def test_composed_backends_fail_before_sampling_with_specific_tsim_guidance():
```
jasonhan3, line 78: I think this test checks that the other backends do indeed import tsim upon sampling; that might not be the behavior that we want.

python/tests/test_device.py
```diff
@@ -215,6 +230,35 @@ def test_noiseless_tsim_circuit_compiles_samplers():
     assert noiseless.compile_detector_sampler() is not None
 
 
+@pytest.mark.parametrize(
+    "backend",
+    [
+        pytest.param(TsimSimulatorBackend(), id="tsim"),
```
jasonhan3, line 236: it's a little strange that TsimSimulatorBackend() doesn't support the seed argument like the other backends. But if we go with design decision 7. to get rid of the "seed" field on the class and instead make it an argument on run(), then this won't be a feature anymore and thus doesn't need to be tested.

python/tests/test_device.py
```diff
@@ -310,70 +352,151 @@ def bell_pair():
         assert all(isinstance(b, bool) for obs in result.observables for b in obs)
 
 
-def _mock_task(*, is_clifford: bool) -> MagicMock:
```
jasonhan3, line 313: are we skipping the is_clifford test? We can still have it for TsimSimulatorBackend

python/tests/test_device.py
```diff
@@ -310,70 +352,151 @@ def bell_pair():
...
+    GeminiLogicalSimulatorTask.run(task, shots=1, with_noise=False)
 
-    task.measurement_sampler.sample.assert_called_once_with(shots=1)
-    task.tsim_circuit.stim_circuit.compile_sampler.assert_not_called()
+    task.simulator.backend.sample.assert_called_once_with(
+        "noiseless-kernel", shots=1, run_detectors=False
+    )
 
 
-def test_run_detectors_clifford_converts_via_m2d():
```
jasonhan3, line 346: not sure why we got rid of this test to replace it with the below different one that tests run_detectors=True

python/tests/test_device.py
jasonhan3: not sure why we got rid of this test to replace it with the below different one that tests run_detectors=False

demo/physical_arch_customization.py
```diff
@@ -373,13 +373,13 @@ def main():
 # We have implemented some basic circuit optimization through our ASAP and ALAP gate scheduling. These passes are classes that you can tell the compiler to use.
 
 # %%
-physical_sim_task_asap = simulator.task(main, place_opt_type=ASAPPlacePass)
+physical_sim_task_asap = PhysicalSimulator(place_opt_type=ASAPPlacePass).task(main)
```
jasonhan3, line 376: Because we are doing compilation on server-side, I think it makes sense to have the compiler options in the constructor of the simulator; however it might be good to construct some kind of "configuration" option that stores the compiler configurations that PhysicalSimulator ingests.

maybe such encapsulation can be done in a future PR

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -207,9 +194,8 @@ class PhysicalSimulatorTask(Generic[RetType]):
     """The physical move kernel compiled from the source SQuIn kernel."""
     _post_processing: atom.PostProcessing[RetType] = field(repr=False)
     """The post-processing object for extracting detectors, observables, and return values."""
-    _thread_pool_executor: ThreadPoolExecutor = field(
-        default_factory=ThreadPoolExecutor, init=False
-    )
+    simulator: GeminiPhysicalSimulator = field(repr=False)
```
jasonhan3, line 197: again, does the SimulatorTask need to store the entire simulator as a field?

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -230,212 +216,29 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
...
     noise_model: NoiseModelABC = field(default_factory=_default_noise_model)
     """The physical noise model used for simulation."""
     arch_spec: ArchSpec = field(default_factory=_default_arch_spec)
     """The physical architecture specification used for compilation."""
+    backend: AbstractSimulatorBackend = field(default_factory=TsimSimulatorBackend)
+    """Sampling and detector-model backend used by created tasks."""
+    place_opt_type: type[passes.Pass] | None = None
+    """Optional placement pass used for compilation."""
+    placement_strategy: PlacementStrategyABC | None = None
+    """Optional physical placement strategy."""
```
jasonhan3, line 233: maybe encapsulate these into a class?

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -230,212 +216,29 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
...
+    m2dets: list[list[int]] | None = None
+    """Optional measurement-to-detector matrix."""
+    m2obs: list[list[int]] | None = None
+    """Optional measurement-to-observable matrix."""
```
jasonhan3, line 237: m2dets/m2obs should not go in the constructor but can maybe go in a separate method we define; see "decisions we've made" 1.

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -537,6 +340,16 @@ def run_async(
...
     ) -> Future[DetectorResult]: ...
 
+    @overload
+    def run_async(
+        self,
+        physical_kernel: ir.Method[[], RetType],
+        shots: int = 1,
+        with_noise: bool = True,
+        *,
+        run_detectors: bool,
```
jasonhan3, line 350: similar comment to logical simulator -- what's the point of supplying run_detectors: bool here if the point of the override is to specify a concrete boolean to dictate the return type?

jasonhan3 (changes requested):
I looked through the commits and have a variety of comments; most of them are summarized in the design decisions comment. Due to the volume of the comments, I think it would be easiest to have a meeting to decide on the design decisions, and from there we can iterate on another implementation.

jasonhan3:
Closed in favor of merging https://github.com/QuEraComputing/bloqade-lanes/pull/820

PR #798, opened 2026-07-10
chore(benchmarks): refresh physical + logical baselines for #788 (always_merge)

weinbe58:
## Why

`main`'s `benchmarks (physical)` / `benchmarks (logical)` CI jobs are **red** since #788 merged. Those jobs run `just benchmark-{physical,logical}`, which `--compare`s current output against the committed baseline CSVs and fails on any drift in the deterministic columns (`move_count_events`, `move_count_lanes`, `estimated_fidelity`, `nodes_explored`).

#788 (`merge all StaticPlacement so no-return placement threads state across CZ layers`) changed those metrics — the merged CZ blocks give the entropy search more to explore — but the baselines weren't regenerated, so the comparison gate now fails. The failing log, e.g.:

```
- ghz_6/rust_entropy_10/logical nodes_explored: baseline=8 current=55 (degraded)
```

## What

Regenerated both baselines on current `main` (which includes #788):

```
uv run python -m benchmarks.cli --architecture physical
uv run python -m benchmarks.cli --architecture logical
```

Structural changes (all consistent with `always_merge`):
- **logical** (6 rows): `ghz_6` entropy `nodes 8→55`; `steane_logical_5` entropy `53→54`.
- **physical** (15 rows): mostly `rust_entropy_*` `nodes_explored` shifts (`bv_70` `622→914`, `adder_64` up, `trotter_rand_35`), plus `move_count` changes where the merged placement reroutes (`steane_physical_35`, `trotter_rand_35`).

`move_count`/`fidelity` are otherwise stable; the regenerated `ghz_6/rust_entropy` value (55) matches what CI computed on the merged fix. `wall_time_ms` also changes but is machine-local and not gated by the comparison.

Regenerated from a clean worktree off current `main`.


weinbe58:
Superseded by #799, which flips `block_spectators` to default `True` and carries its own refreshed benchmark baselines matching that default. This PR's baselines were regenerated for `block_spectators=False` (nodes=55), which no longer matches main's behavior once #799 lands. Closing in favor of #799.

PR #799, opened 2026-07-10
feat(lanes): scope CZ solves via block_spectators (default on) + Palindrome lookahead knob

weinbe58:
First slice of decomposing #683 (user-directed movement) — two placement knobs that are **independent of the movement dialect**, lifted onto `main`.

## Context

#788's `always_merge` merges all qubits into one `StaticPlacement` block, so each CZ solve searches over *every* qubit in the block (not just the CZ's two participants), inflating `nodes_explored` — even though unrelated qubits never move.

## Changes

- **`RustPlacementTraversal.block_spectators` (default `True`)** — qubits not in the CZ's `controls`/`targets` are treated as blocked obstacles for that solve, so search effort no longer scales with unrelated qubits in a merged block. The goal layout backfills spectators from the current state (`goal_map.get(qid, state.layout[qid])`), since a scoped solve only returns participants' goal config (a no-op when off).
- **`PalindromePlacementStrategy.lookahead` (default `1`)** — forward only the leading N CZ layers to the inner solver. Under palindrome, atoms return home before each CZ, so future layers can't inform the current placement.
- **Refreshed benchmark baselines** to match the new default.

## Why default `True` (net win on the current suite)

Measured `True` vs `False` on current `main`:
- **Logical:** `ghz_6` entropy nodes **55→8**; `steane_logical_5` 54→53; moves & fidelity unchanged.
- **Physical:** `bv_70` entropy nodes **914→622**; `adder_64` down; `steane_physical_35` entropy gets **fewer moves (134→98 events) *and* higher fidelity (0.0025→0.0099)**; `trotter_rand_35` roughly flat, moves equal-or-fewer.
- Move counts are equal-or-fewer and fidelity equal-or-better across the changed rows; deterministic strategies (astar/bfs/dfs/greedy/ids) are unchanged. The move-count regression the feature branch once cited (`c11e6d74`, DFS `trotter_rand_35` +20) does **not** reproduce on current `main` with the completed backfill.

## Verification

- Benchmark comparison (logical): **no differences** against the refreshed baselines.
- **Full suite: 1556 passed / 0 failures**; black/ruff/pyright clean.

## Supersedes #798

#798 refreshed baselines for `block_spectators=False` (nodes=55), which no longer matches `main`'s default once this lands, so it has been **closed** in favor of this PR (which carries its own refreshed baselines matching `block_spectators=True`).


weinbe58:
Thanks @copilot — responses to the two review comments:

- **`strategy.py` (lookahead):** addressed in 0b36f496 — added a `lookahead >= 0` validation and converted the no-op string literal (a dead expression statement inside `__init__`, not an attribute docstring) into a real comment.
- **`movement.py` (`traced_blocked_locations` vs the scoped blocked set):** valid, and it's now the default so the trace under-reports for scoped solves. Rather than mirror the blocked set in Python, the fix belongs in the backend (the trace should carry `SearchContext.blocked`). Tracked in #800.

Issue #801, opened 2026-07-11
Clean up top-level bloqade.lanes modules: consolidate compile entry points + relocate Gemini-specific APIs


weinbe58:
## Context

`bloqade.lanes/__init__` was just emptied of eager re-exports to break a `lanes ↔ gemini` import cycle (lanes submodules like `rewrite.circuit2place` import `bloqade.gemini.logical`, while `gemini` imports `bloqade.lanes`; a re-export re-entered an in-progress gemini import — see the movement-dialect work). That exposed how tangled/duplicated the top-level module surface is.

**Layering principle:** `bloqade.lanes` = machine-agnostic movement compilation (lower layer); `bloqade.gemini` = Gemini-machine specifics (higher layer, builds on lanes). Anything Gemini-specific should live in `gemini`; `lanes` must not import upward.

## Goals
1. Remove stale/duplicate APIs and consolidate old→new.
2. Relocate Gemini-specific APIs into `bloqade.gemini` to prevent re-coupling.

Do it as **one** change (checkboxed workstreams below).

## Inventory (top-level `bloqade/lanes/*.py`) — importers = lanes/gemini/tests/demos

| Module | Disposition |
| --- | --- |
| `types.py`, `utils.py`, `transform.py`, `passes.py`, `noise_model.py` | **keep** (core / cross-cutting; `noise_model` is consumed by lanes internals so cannot move) |
| `upstream.py` (`squin_to_move`, `NativeToPlace`, `PlaceToMove`) | **keep in lanes** — the low-level, machine-agnostic squin→move compile. (Optional: rename; "upstream" is opaque.) |
| `_prelude.py` (structural; the live one — 7 tests, 2 demos) | **keep**; consider promoting to public `prelude` |
| `prelude.py` (basic; legacy) | **remove** — sole user is `rewrite/transversal.py`; migrate it to `_prelude` |
| `compile.py` + `logical_mvp.py` | **consolidate + relocate to `bloqade.gemini`** (see WS1) |
| `steane_defaults.py` (`steane7_m2*`) | **relocate → `bloqade.gemini`** (Steane [[7,1,3]] is Gemini/logical-specific; only `logical_mvp` uses it) |
| `cudaq_integration.py` | **relocate → `bloqade.gemini`** (hardware-specific; only `logical_mvp` uses it) |
[...]

PR #802, opened 2026-07-11
feat: explicit user-directed atom movement dialect (movement_kernel + move_to)

### LINKED ISSUE #669: Add a User-Directed Atom Movement Dialect Composable with Squin
# Add a User-Directed Atom Movement Dialect Composable with Squin

## Summary

Add a new IR dialect for user-directed atom movement intent, rather than adding movement-placement statements directly to the Squin dialect.

This dialect would provide a statement that lets users request that selected qubits be moved to specified physical `LocationAddress` destinations. The new dialect should be composable with Squin by defining a new dialect group, for example:

```python
movement_squin = squin.kernel.union([movement_intent])
```

or a similar Bloqade/Kirin dialect-group construction.

The goal is to let users write kernels that combine normal Squin circuit operations with explicit intermediate atom-placement requests, while keeping Squin itself focused on quantum circuit semantics.

## Motivation

Today, physical movement decisions are mostly compiler-driven. Users can influence initial layout, but they cannot directly express intermediate movement intent inside a kernel.

This makes it difficult to:

- prototype hand-designed movement schedules
- debug AOD/SLM routing behavior
- express hardware-aware algorithms where useful intermediate atom positions are known
- compare automatic placement against user-guided placement

Adding this directly to Squin would blur the boundary between circuit semantics and physical movement intent. A separate dialect is cleaner: users opt into the extra physical-control surface by compiling with a composed dialect group.

## Proposed Design

Create a new dialect, tentatively named one of:

```text
movement_intent
physical_hint
placement_hint
atom_move_hint
```

The dialect should define a statement with semantics like:

```python
move_to(qubits, locations)
```

where:

- `qubits` is an `IList[Qubit, N]`
- `locations` is an `IList[LocationAddress, N]`
- each qubit is requested to move to the corresponding location

Example use in a composed kernel group:

```python
@movement_squin.kernel
def kernel():
    q = qubit.qalloc(4)

    movement.move_to(
        [q[0], q[1]],
        [
            LocationAddress(word_id=4, site_id=0),
            LocationAddress(word_id=4, site_id=2),
        ],
    )

    squin.cz(q[0], q[2])
```

The exact namespace is open. Possible API shapes:

```python
movement.move_to(qubits, locations)
movement.place_at(qubits, locations)
movement.request_move(qubits, locations)
```

## Dialect Group Composition

The movement-intent dialect should not be part of the base Squin dialect group by default.

Instead, Bloqade Lanes should expose a composed dialect group for kernels that need this feature, such as:

```python
from bloqade import squin
from bloqade.lanes import movement

@movement.kernel
def kernel():
    ...
```

or:

```python
from bloqade.lanes.groups import physical_squin

@physical_squin.kernel
def kernel():
    ...
```

The final API naming is an open design question, but the important architectural requirement is:

```text
Squin remains circuit-level.
The new dialect carries physical movement intent.
Users opt in by compiling with a dialect group that includes both.
```

## Compiler Behavior

At the new movement-intent dialect level, the statement represents a user-directed intermediate placement request.

When lowering to the place dialect, the compiler should encode:

- the selected qubits
- their current positions
- the requested destination positions
- enough ordering information to preserve where the request occurs relative to nearby circuit operations

Then, during place-to-move lowering, the existing move synthesizer should synthesize legal AOD moves to satisfy the requested placement.

For each CZ stage, one possible behavior is:

1. Synthesize moves for any user-requested intermediate placements that occur before the CZ.
2. Check which CZ pairs are already physically paired.
3. For remaining unpaired CZ pairs, synthesize additional compiler-chosen moves as today.
4. Emit the CZ pulse.
5. Apply palindrome/return behavior if enabled.

## Interaction With CZ Placement

User-directed movement should act as a placement constraint or hint before subsequent operations.

If a user-directed move places some qubits into CZ-compatible positions, the CZ placement stage should recognize those pairs as already paired.

For pairs that are not yet CZ-compatible, the existing placement/move synthesis logic should still be allowed to move additional atoms.

Open design question:

- Should `move_to` be a hard constraint that must be satisfied exactly?
- Or should there be both hard and soft forms, such as `move_to` and `prefer_move_to`?

## Palindrome / Return Moves

A tricky part is interaction with palindrome movement.

Today, palindrome movement effectively undoes compiler-generated moves after a CZ pulse. With user-directed intermediate moves, we need to decide what should be undone.

One possible design:

- Track all movement segments inserted between CZ pulses.
- Include user-directed moves in that movement segment.
- After the CZ pulse, undo the full segment when palindrome behavior is enabled.
- Treat user-directed moves as temporary unless explicitly marked persistent.

Potential future extension:

```python
movement.move_to(qubits, locations, persistent=True)
```

where `persistent=True` means the user wants the new positions to become the post-CZ current positions rather than being palindromed away.

Open questions:

- Are user-directed moves temporary by default?
- Should persistent placement be supported in the initial implementation?
- Should palindrome behavior be controlled at the placement-strategy level or the movement-intent statement level?

## Parallelism and AOD Constraints

The user may pass a list of qubits that they conceptually want moved together, but the requested moves may not be physically parallelizable under AOD constraints.

The compiler should not assume the list implies one parallel AOD shot.

Instead, the move synthesizer should:

- attempt to pack compatible moves into as few AOD shots as possible
- split incompatible moves across multiple shots
- preserve correctness over user-implied parallelism
- optionally emit diagnostics when the requested movement cannot be performed in one AOD shot

Potential future extension:

```python
movement.move_to(qubits, locations, require_parallel=True)
```

If `require_parallel=True`, compilation would fail if the requested moves cannot be scheduled in a single AOD shot.

## Validation / Edge Cases

The implementation should validate:

- `len(qubits) == len(locations)`
- each destination is a valid `LocationAddress`
- destination locations exist in the target architecture
- no two qubits are assigned the same destination
- destination is not occupied by an unmoved atom unless that atom is also part of a valid simultaneous/permuted move
- requested movement is physically reachable
- requested movement does not violate AOD lane constraints
- behavior is well-defined if a requested qubit is already at the requested location

Invalid requests should produce clear compiler errors.

## Suggested Lowering Path

Proposed flow:

```text
Squin + movement-intent composed dialect group
    ↓
movement-intent move_to statement
    ↓
Place dialect statement or placement constraint encoding requested positions
    ↓
Place-to-move lowering
    ↓
Move synthesizer generates legal AOD Move statements
```

The implementation should reuse the existing move synthesis machinery rather than requiring users to construct low-level Move IR directly.

## Non-Goals

- Do not add this statement directly to the core Squin dialect.
- Do not require all Squin users to compile with movement-intent semantics.
- Do not expose raw low-level Move IR construction as the primary user API.
- Do not guarantee that all qubits passed to one `move_to` call move in one AOD shot unless an explicit `require_parallel` option is added.

## Acceptance Criteria

- A new movement-intent dialect exists.
- The movement-intent dialect defines a statement for requesting qubits be moved to destination `LocationAddress` values.
- A composed dialect group exists that allows users to write kernels containing both Squin operations and movement-intent statements.
- The new statement accepts lists of qubits and matching destination locations.
- Lowering to the place dialect preserves the requested placement information.
- Place-to-move lowering uses the existing move synthesizer to emit legal Move IR.
- Invalid locations and mismatched list lengths produce actionable errors.
- Movement requests interact correctly with CZ placement synthesis.
- Palindrome behavior is documented and covered by tests.
- Tests cover:
  - valid single-qubit movement request
  - valid multi-qubit movement request
  - mismatched qubit/location lengths
  - invalid destination location
  - duplicate destination locations
  - movement request followed by CZ
  - movement request under palindrome placement strategy
  - movement request requiring multiple AOD shots
  - composed Squin + movement-intent dialect group compilation

## Open Questions

- What should the dialect be named?
- What should the composed dialect group be named?
- Should user-directed moves be hard constraints or hints?
- Should user-directed moves be palindromed by default, or should they persist?
- Should the API expose a `persistent` option?
- Should the API expose a `require_parallel` option?
- Should invalid-but-nearby requests fail immediately, or should the compiler attempt best-effort routing?
- Should diagnostics report why requested moves could not be parallelized?
- Should the statement live under `bloqade.lanes`, `bloqade.gemini`, or another namespace?

### LINKED ISSUE #784: Add abstraction layer for word/site ID's to index positions on hardware, invariant of architecture
@bachbao noticed a breaking change where, with the new physical architecture specification for Gemini, his old word/site ID's no longer worked.

It might be worth introducing an abstraction layer that fixes the word/site ID's that a user can allocate qubits on, and the compiler can map the user-intended atom positions to the internal word/site ID's for the architecture specification, to decouple the user's intention of atom locations with the internal representation of the word/site ID's on the architecture specification (so that if the architecture specification changes in the future, the user does not have to change their code).

weinbe58:
Explicit user-directed atom movement for the physical compiler: an `arrange` dialect (`move_to` / `permute`) plus a companion `arch` dialect (`loc` / `cz_partner` / location attributes), lowered through place → move, with the placement analysis that tracks user moves. Reworked cleanly on top of current `main` — **supersedes #683** (same feature, rebased past #788/#799/#801, no design docs).

closes #784

### What's in it

**`arrange` dialect (`gemini.common.arrange`)** — user-facing movement primitives:
- `move_to(qubits, locations)` — move qubits to explicit `LocationAddress` destinations within a zone; the compiler tracks where the atoms end up.
- `permute(qubits, perm, insert_moves=False)` — a logical relabel where `qubits[i]` comes to refer to what was `qubits[perm[i]]`. **Relabels by default** (free — the quantum information is permuted and a later transversal move absorbs the physical permutation lazily); `insert_moves=True` also **commits** the physical permutation now (rejected under `PalindromePlacementStrategy`, which would return the committed moves home).

**`arch` dialect (`lanes.arch`)** — architecture-resolved statements, const-folded once the pipeline binds an `ArchSpec`:
- `loc(zone, row, col)` — resolve a grid coordinate to a physical `LocationAddress` via the new `ArchSpec.location_at`.
- `cz_partner(loc)` — resolve the CZ blockade partner.
- `word_id` / `site_id` / `zone_id` — attribute reads on a `LocationAddress`.
- An `ArchResolvedStmt` base carries the optional `arch_spec`; a single `BindArchSpec` pass binds it before folding. Unresolvable coordinates / partners raise `InterpreterError` during analysis (caught under `run_no_raise`; surfaced under `no_raise=False`), consistent across const-prop, the interpreter, and `move_to` validation.

**Lowering + placement analysis** — `place.MoveTo` / `place.Permute`, `circuit2place` / `place2move` lowering, reorder-around-movement, pipeline/upstream wiring. `UserMoved` / `Permuted` lattice states and `MoveToPlacementStrategyABC.{move_to_placements, permute_placements}`. `measure_placements` is a concrete `PlacementStrategyABC` method (canonical-layout `ExecuteMeasure`, shared by all strategies).

**Demo** — `demo/move_demo.py`: an [[8,3,2]] logical-gadget demo (transversal CX, logical CCZ, logical SWAP via `permute`) compiled end-to-end through `PhysicalPipeline`, with a `VIRTUAL` toggle selecting relabel-only vs committed-move permutation.

### Breaking changes
- `bloqade.lanes` no longer re-exports device classes, `Metrics`, noise-model helpers, or `steane7_*` at the package root — import them from their submodules / `bloqade.gemini`. (Broader top-level cleanup tracked in **#801**.)
- `arch.loc` takes `(zone, row, col)` grid coordinates, not `(zone_id, word_id, site_id)`.
- The dialect is named `arrange` (renamed from `movement`); `permute` relabels by default (`insert_moves` commits the moves).

### Notable fix
`HoistNewQubitsUp` / `MergeStaticPlacement` dropped `place.Permute.insert_moves` when rebuilding merged `StaticPlacement` blocks, so a `permute` compiled through the ASAP/ALAP place passes silently lost the flag (a `KeyError` under `no_raise=False`, a silent no-op under the default `no_raise=True`). Fixed at both reconstruction sites + regression-tested against `ASAPPlacePass`.

### Compatibility
Preserves behaviour already on `main`: merged-`StaticPlacement` no-return correctness (#788), `block_spectators` (default on) + Palindrome `lookahead` (#799), and #788's canonical-layout / `EndMeasure.qubits` measurement ordering.

### Verification
- Full Python suite: **1665 passed, 6 skipped**.
- Rust core tests: **213 passed**.
- Lint green: ruff, black, isort, pyright; cargo fmt, clippy.


python/bloqade/lanes/rewrite/circuit2place.py
```diff
@@ -349,6 +352,75 @@ def rewrite_StarRz(self, node: gemini_stmts.StarRz) -> abc.RewriteResult:
...
+
+        if len(location_attrs) != len(inputs):
+            # Fail loudly: a mismatched move_to would otherwise be silently
+            # miscompiled (move_to_placements zips qubits/locations and drops
+            # extras). MoveToValidation catches this too, but it is skipped
+            # under verify=False, so guard here as well.
+            raise ValueError(
+                f"move_to: number of locations ({len(location_attrs)}) does not "
+                f"match number of qubits ({len(inputs)})"
+            )
```
weinbe58, line 377: please make this an assertion. This should be caught by typeinference.

python/bloqade/lanes/rewrite/circuit2place.py
```diff
@@ -349,6 +352,75 @@ def rewrite_StarRz(self, node: gemini_stmts.StarRz) -> abc.RewriteResult:
...
+        perm_attr = tuple(int(p) for p in perm_hint.data)
+
+        if sorted(perm_attr) != list(range(len(inputs))):
+            # Fail loudly: perm must be a permutation of range(len(qubits)); a
+            # bad perm would otherwise IndexError deep in placement or silently
+            # miscompile. Guard here since validation may be skipped (verify=False).
+            raise ValueError(
+                f"permute: perm {perm_attr} is not a permutation of "
+                f"range({len(inputs)})"
+            )
```
weinbe58, line 411: make this an assertion, this shuold be caught by type inference.

weinbe58:
@WanHsuanLin we will schedule some time to review this together so don't worry if you are not sure about the details

python/bloqade/lanes/analysis/placement/strategy.py
```diff
@@ -61,25 +78,224 @@ def sq_placements(
...
+        are measured there, and the compiler tracks the final positions within
+        the single zone. ``UserMoved`` is a ``ConcreteState`` subclass, so it
+        flows through the concrete-state path below. ``PalindromePlacementStrategy``
+        overrides this to reject ``UserMoved`` instead: under palindrome a
+        user-move is only committed by a following CZ, never a measurement.
+        """
+        if not isinstance(state, ConcreteState):
+            return state
+        if len(qubits) != len(state.layout):
+            return AtomState.bottom()  # all qubits must be measured
```
weinbe58: please make this an assertion

python/bloqade/gemini/physical/group.py
```diff
@@ -0,0 +1,126 @@
...
+from kirin.validation import ValidationSuite
+from typing_extensions import Doc
+
+from bloqade.gemini import common as gemini_common
+from bloqade.gemini.logical.group import kernel as logical_kernel
+from bloqade.lanes.arch.spec import ArchSpec
+from bloqade.lanes.dialects import arch as arch_dialect
+
+
+@ir.dialect_group(logical_kernel.union([gemini_common.dialects.arrange, arch_dialect]))
```
weinbe58, line 19: look into this composition

PR #805, opened 2026-07-13
Use AddressAnalysis for _find_qubit_ssas

### LINKED ISSUE #803: Potential bug introduced by #773
copilot was correct here: https://github.com/QuEraComputing/bloqade-lanes/pull/773/changes#r3530846611

The way to fix this is to either add a trait upstream to qubit allocations or to combine this with AddressAnalysis to make sure you map the SSA value to a particular qubit allocation index.

jasonhan3:
closes https://github.com/QuEraComputing/bloqade-lanes/issues/803

# Use Address Analysis for Qubit SSA Discovery

## Summary

Replace the type- and statement-class-based implementations of
`_find_qubit_ssas` with `AddressAnalysis` in both the physical simulator and
logical compiler paths.

The helpers now return one representative SSA value for each concrete qubit
allocation address, in allocation-address order.

## Motivation

The previous implementations inferred allocations from IR surface structure:

- The physical helper collected every result typed as `QubitType`.
- The logical helper collected only `qubit.New` statements.

Neither rule represents the intended invariant: count each allocated qubit
once, even when that qubit has multiple SSA aliases. The logical rule also
missed Gemini `new_at(...)` allocations entirely.

`AddressAnalysis` already models the allocation semantics of both
`qubit.new()` and Gemini `new_at(...)`, while propagating concrete addresses
through lists, constant indexing, aliases, helper calls, and supported control
flow.

## Changes

- Run `AddressAnalysis` in both `_find_qubit_ssas` helpers.
- Collect the first SSA value associated with each `AddressQubit` address.
- Return those values ordered by address (`0, 1, ..., qubit_count - 1`), which
  corresponds to allocation order.
- Document that `qalloc` must be aggressively unrolled before an SSA-returning
  helper can recover its individual allocation results from the caller IR.
- Add regression coverage that verifies logical `new_at(...)` allocations are
  included.
- Update physical coverage so arbitrary `QubitType`-producing statements with
  no concrete address-analysis handler are not treated as allocations.

## Behavioral Notes

This intentionally changes the interpretation of aliases. For example,
multiple SSA values referring to `reg[0]` now contribute one allocation rather
than multiple qubits.

`AddressAnalysis.qubit_count` can count allocations in an unexpanded
`qalloc(3)` call by analyzing the helper body. However, the SSA-returning
helpers require aggressive unrolling: before unrolling, the caller contains a
single register SSA rather than the three individual `qubit.new()` result
SSAs.

The physical annotation path already aggressively unrolls before calling its
helper. Logical kernels using `qalloc` must likewise be aggressively unrolled,
as they were for the previous helper to find `qubit.New` statements.

## Tests

- `uv run pytest python/tests/gemini/test_physical_simulator.py`
- `uv run pytest python/tests/test_cudaq_integration.py`
- `uv run ruff check python/bloqade/gemini/device/physical_simulator.py python/bloqade/lanes/logical_mvp.py python/tests/gemini/test_physical_simulator.py python/tests/test_cudaq_integration.py`
- `uv run black --check python/bloqade/gemini/device/physical_simulator.py python/bloqade/lanes/logical_mvp.py python/tests/gemini/test_physical_simulator.py python/tests/test_cudaq_integration.py`
- `uv run pyright python/bloqade/gemini/device/physical_simulator.py python/bloqade/lanes/logical_mvp.py python/tests/gemini/test_physical_simulator.py python/tests/test_cudaq_integration.py`

## Compatibility

No public API changes. The private helpers now provide canonical allocation
semantics instead of returning arbitrary qubit-typed SSA aliases.

### DIFF
--- a/python/bloqade/gemini/device/physical_simulator.py
+++ b/python/bloqade/gemini/device/physical_simulator.py
@@ -12,13 +12,14 @@ from typing import (
 )
 
 import numpy as np
+from bloqade.analysis.address import AddressAnalysis, AddressQubit
 from bloqade.analysis.fidelity import FidelityAnalysis
 from bloqade.decoders.dialects.annotate.stmts import SetDetector, SetObservable
 from bloqade.rewrite.passes import AggressiveUnroll
-from kirin import ir, passes, rewrite, types
+from kirin import ir, passes, rewrite
 from kirin.dialects import func, ilist, py
 
-from bloqade import qubit, types as bloqade_types
+from bloqade import qubit
 
 from .simulator import DetectorResult, Result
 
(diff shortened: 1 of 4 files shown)


weinbe58 (approved):
LGTM

PR #809, opened 2026-07-15
feat(python): add MoveToStackMove transformation

weinbe58:
## Summary

Adds a `MoveToStackMove` transformation to `bloqade.lanes.transform`, in the same style as the existing `MoveToSquin*` objects. It wires the previously standalone move → stack_move rewrite pieces (`RewriteMoveToStackMove`, `stackify`, `dump_program`) into a single orchestrator.

Before this change these pieces existed but were only ever composed manually in unit tests — there was no `transform`-style entry point that lowered a `move`-dialect kernel to a canonicalized `stack_move` kernel.

## What it does

`MoveToStackMove.emit(main, no_raise=True) -> ir.Method` runs the full lowering pipeline on a **copy** of the input (input is left untouched):

1. `RewriteMoveToStackMove` — move → stack_move (strips Load/Store state threading, materializes address attributes as `Const*` SSA values, rebuilds Measure/AwaitMeasure/GetItem).
2. DCE + CSE to a fixpoint — the cleanup the encoder pipeline expects before stackification.
3. `similar(discard(move.dialect))` — drop the now-unused source dialect.
4. `stackify` — normalize into stack-consistent form (single block, each SSA value used at most once, defining statements in stack order).
5. `verify()` gated on `no_raise` (skips `verify_type`/`TypeInfer`, which `stack_move` has no rules for).

`emit_bytecode(main, version=(1, 0), no_raise=True) -> Program` additionally encodes the result to a bytecode `Program` via `dump_program`.

## Tests

New `python/tests/test_transform_move_to_stack_move.py` (6 tests): lowering correctness, stack-consistency, input immutability, the `no_raise=False` verify path, bytecode round-trip, and version passthrough. All pass alongside the existing `test_move2stack_move` / `test_stackify` suites (36 total). `isort` / `black` / `ruff` / `pyright` all clean.

### DIFF
--- a/python/bloqade/lanes/transform.py
+++ b/python/bloqade/lanes/transform.py
@@ -185,11 +185,11 @@ class MoveToStackMove:
        each SSA value used at most once, defining statements in stack order).
 
     ``RewriteMoveToStackMove`` only lowers the subset of ``move`` statements
-    the bytecode path supports, so after dropping ``move`` from the dialect
-    group ``emit`` runs ``statements_outside_dialect_group`` and raises if any
-    statement is left outside the group — Kirin's ``verify()`` does not check
-    dialect-group membership, so an unlowered statement would otherwise slip
-    through and fail lazily inside ``dump_program``.
+    the bytecode path supports. When ``no_raise`` is ``False``, ``emit`` runs
+    ``statements_outside_dialect_group`` after dropping ``move`` from the group
+    and raises if any statement is left outside it — Kirin's ``verify()`` does
+    not check dialect-group membership, so an unlowered statement would
+    otherwise slip through and fail lazily inside ``dump_program``.
 
     ``emit_bytecode`` runs ``emit`` and encodes the result to a bytecode
     ``Program`` via ``dump_program``.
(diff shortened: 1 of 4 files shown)


python/bloqade/lanes/transform.py
```diff
@@ -162,3 +167,82 @@ def _get_noise_model(self) -> NoiseModelABC | None:
...
+        # so an unlowered statement would otherwise surface as a confusing
+        # EncodingError deep inside dump_program.
+        leftover = statements_outside_dialect_group(out)
+        if leftover:
+            kinds = sorted({type(stmt).__name__ for stmt in leftover})
+            raise ValueError(
+                "MoveToStackMove left statements outside the stack_move dialect "
+                f"group: {', '.join(kinds)}; RewriteMoveToStackMove does not "
+                "lower them, so the kernel cannot be emitted as stack_move IR"
+            )
```
weinbe58: Can you put this under no-raise? flag

weinbe58: Done in 02d55bf7 — the `statements_outside_dialect_group` check now runs only when `no_raise=False` (alongside `verify()`). With `no_raise=True` emit stays best-effort and silent. Updated the docstring and the fail-fast test accordingly.

PR #814, opened 2026-07-16
fix(noise): add move + correlated CZ noise to Steane init (#796)

### LINKED ISSUE #796: More faithful noise model for Steane state prep circuit
The current noise model for state prep does not appropriately represent the predicted hardware noise on Gemini:
1. It is missing move noise
2. It is missing a correlated CZ noise channel, as is present in the rest of the circuit

Image

e.g. this is a transversal gate, which does include move and correlated noise:

Image

Can we please add more noise and a more faithful CZ channel?

Note: this issue risks being a blocker for the Gemini Benchmarking project, as a lot of the benchmarking is characterizing various state preps.

weinbe58:
## Summary

Closes #796.

The noisy Steane [[7,1,3]] initialization kernel did not faithfully represent the predicted Gemini hardware noise for state prep. Missing/incorrect sources have been added:

1. **Move noise on both legs** — mover/sitter single-qubit Pauli channels (+ loss) were only applied *after* each CZ. They are now also applied *before* each CZ, representing both the move-in and move-out legs of a transversal CZ.
2. **Correlated CZ noise channel** — a `two_qubit_pauli_channel` is now emitted after every CZ layer, carrying the noise model's paired error probabilities (in `PAIRED_KEYS` order), mirroring the physical `cz_paired_noise` kernel.
3. **CZ paired/unpaired loss** — new `cz_paired_loss` / `cz_unpaired_loss` args apply post-CZ loss to CZ participants and non-participating columns respectively, matching the physical noise model.
4. **Bystander (rest) sitter noise** — columns not participating in a given CZ layer now receive sitter noise, since they idle through the move.
5. **Layer 2 mover/sitter assignment fix** — the `cz_w_noise` helper had the mover/sitter roles inverted for CZ layer 2 (`evens[:-1]` was receiving sitter noise and `[3,5,6]` was receiving mover noise). Renamed the helper's parameters to `(mover_ids, sitter_ids)` so the intent is self-documenting, and reordered all three call sites.

`generate_logical_noise_model` threads the scaled `two_qubit_pauli` error probabilities into `steane7_initialize_with_noise` via a new `cz_errors` argument (defaults to zeros when omitted) and forwards the CZ loss probabilities.

## Testing

- `python/tests/test_noise_init_generator.py` — 10 tests including: one correlated CZ channel per CZ layer (none in the clean kernel), and the channel carries the model's paired probabilities.
- Full related suite (noise / logical / steane): 135 tests pass.
- Both clean and noisy kernels specialize **and** verify cleanly through `SpecializeGemini(no_raise=False)`, and the noisy kernel now verifies through `MoveToSquinLogical(aggressive_unroll=True)` (fixed by seeding the `get_rows` foldl accumulator with `ilist.IList([], elem=qubit.QubitType)` so the length TypeVar propagates to CZ operands).
- `python/benchmarks/harness/latest_logical.csv` — regenerated to reflect the new noise model (per-case fidelity update, timings/moves/nodes unchanged):
  - `ghz_4`:            `0.7306791625` → `0.4261944432`
  - `ghz_6`:            `0.6354523319` → `0.2830769898`
  - `steane_logical_5`: `0.7053272104` → `0.3595356539`
- black / isort / ruff / pyright all clean.


python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -213,7 +223,16 @@ def _inner(cumulant, i):
             squin.broadcast.qubit_loss(local_loss_prob, cols0_5)
 
         # CZ layer 1: controls=odds (sitters), targets=evens[1:] (movers)
+        squin.broadcast.single_qubit_pauli_channel(
```
jasonhan3: Just a comment: so in this kernel we are assuming that the targets move to the controls?

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -213,7 +223,16 @@ def _inner(cumulant, i):
...
+            sitter_px, sitter_py, sitter_pz, cols_odds
+        )
+        squin.broadcast.single_qubit_pauli_channel(
+            mover_px, mover_py, mover_pz, cols_evens_tail
+        )
         squin.broadcast.cz(cols_odds, cols_evens_tail)
+        squin.broadcast.two_qubit_pauli_channel(
+            cz_errors_ilist, cols_odds, cols_evens_tail
+        )
```
jasonhan3: Are we missing the `cz_unpaired_noise` for qubits that do not have a blockade partner?

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -213,7 +223,16 @@ def _inner(cumulant, i):
             squin.broadcast.qubit_loss(local_loss_prob, cols0_5)
 
         # CZ layer 1: controls=odds (sitters), targets=evens[1:] (movers)
+        squin.broadcast.single_qubit_pauli_channel(
+            sitter_px, sitter_py, sitter_pz, cols_odds
```
jasonhan3: I think all the atoms that are not moved should receive the sitter noise? So not just `cols_odds`, but `cols_odds` and evens[0]?

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -231,7 +250,16 @@ def _inner(cumulant, i):
             squin.broadcast.qubit_loss(local_loss_prob, col6)
 
         # CZ layer 2: controls=evens[:-1] (movers), targets=[3,5,6] (sitters)
+        squin.broadcast.single_qubit_pauli_channel(
+            mover_px, mover_py, mover_pz, cols_evens_head
+        )
+        squin.broadcast.single_qubit_pauli_channel(
+            sitter_px, sitter_py, sitter_pz, cols_356
```
jasonhan3: Same comment; all atoms not moved should receive sitter noise, so we should add `evens[-1]` ?

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -251,7 +279,16 @@ def _inner(cumulant, i):
             squin.broadcast.qubit_loss(local_loss_prob, cols2_6)
 
         # CZ layer 3: controls=evens[:-1] (sitters), targets=odds (movers)
+        squin.broadcast.single_qubit_pauli_channel(
+            sitter_px, sitter_py, sitter_pz, cols_evens_head
```
jasonhan3: Same comment; all atoms not moved should receive sitter noise, so evens[-1] should be added as well

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -213,7 +223,16 @@ def _inner(cumulant, i):
             squin.broadcast.qubit_loss(local_loss_prob, cols0_5)
 
         # CZ layer 1: controls=odds (sitters), targets=evens[1:] (movers)
```
jasonhan3: Do we also assume the same noise model per CZ layer?

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -213,7 +223,16 @@ def _inner(cumulant, i):
...
+        squin.broadcast.single_qubit_pauli_channel(
+            sitter_px, sitter_py, sitter_pz, cols_odds
+        )
+        squin.broadcast.single_qubit_pauli_channel(
+            mover_px, mover_py, mover_pz, cols_evens_tail
+        )
```
jasonhan3: Are we missing the loss channel for the forward move? Maybe we should have something like

```
       if loss:
            squin.broadcast.qubit_loss(sit_loss_prob, cols_odds)
            squin.broadcast.qubit_loss(move_loss_prob, cols_evens_tail)
```

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -213,7 +223,16 @@ def _inner(cumulant, i):
...
         squin.broadcast.cz(cols_odds, cols_evens_tail)
```
jasonhan3: Are we missing the paired CZ loss?

jasonhan3 (changes requested):
Some comments below on the faithfulness of the noise modelling; we can also address these in a future PR (I can make a separate issue for this if we want, for more faithful noise modelling):
1. Do we want the same mover/sitter noise channel for each CZ layer? Each CZ layer might have a different sequence of moves and thus a different set of noise
2. I think all atoms that are not moved should receive the sitter noise?
3. I think we're missing the CZ unpaired noise for qubits that are not in blockade radius  but participate in the CZ gate?
4. Are we missing the qubit_loss channel for the "forward" moves?
5. Are we missing the paired CZ loss channel for the CZ gates?

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -173,117 +199,105 @@ def _get(i: int):
...
+            squin.broadcast.single_qubit_pauli_channel(
+                mover_px, mover_py, mover_pz, movers
+            )
+
+            if loss:
+                squin.broadcast.qubit_loss(sit_loss_prob, sitters)
+                squin.broadcast.qubit_loss(move_loss_prob, movers)
+
+            squin.broadcast.cz(movers, cz_sitters)
+            squin.broadcast.two_qubit_pauli_channel(cz_errors_ilist, movers, cz_sitters)
```
jasonhan3, line 253: Are we missing a single qubit pauli channel for the unpaired atoms in the CZ layer? This is what is done in the physical noise model: https://github.com/QuEraComputing/bloqade-lanes/blob/main/python/bloqade/lanes/noise_model.py#L55-L67

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -173,117 +199,105 @@ def _get(i: int):
...
+            squin.broadcast.single_qubit_pauli_channel(
+                sitter_px, sitter_py, sitter_pz, sitters
+            )
+            squin.broadcast.single_qubit_pauli_channel(
+                mover_px, mover_py, mover_pz, movers
+            )
+            if loss:
+                squin.broadcast.qubit_loss(cz_paired_loss, movers + cz_sitters)
```
jasonhan3: Should this be "correlated_qubit_loss"? https://github.com/QuEraComputing/bloqade-lanes/blob/main/python/bloqade/lanes/noise_model.py#L118

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -173,117 +199,105 @@ def _get(i: int):
...
+                squin.broadcast.qubit_loss(cz_unpaired_loss, others)
```
jasonhan3, line 265: nit: should the CZ loss channels be applied after the CZ layer? Instead of being applied together with the move loss channels

jasonhan3 (changes requested):
Some minor comments about things that are missing; otherwise looks good!

python/bloqade/lanes/arch/gemini/logical/upstream.py
weinbe58: Yes but they should commute because you can't get lost twice

python/bloqade/lanes/noise_model.py
```diff
@@ -214,7 +218,13 @@ def generate_logical_noise_model(
         sitter_px=sitter_px,
         sitter_py=sitter_py,
         sitter_pz=sitter_pz,
+        cz_unpaired_gate_px=noise_model.cz_unpaired_gate_px,
```
jasonhan3, line 221: nit: should the noise channel param here be

```
    cz_unpaired_gate_px, cz_unpaired_gate_py, cz_unpaired_gate_pz = (
        noise_model.cz_unpaired_pauli_rates
    )
```

to use the scaled pauli rates instead of using 

```
noise_model.cz_unpaired_gate_px, noise_model.cz_unpaired_gate_py, noise_model.cz_unpaired_gate_pz
```

? To be consistent with the rest of the code here

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -173,117 +206,114 @@ def _get(i: int):
...
+                sitter_px, sitter_py, sitter_pz, sitters
+            )
+            squin.broadcast.single_qubit_pauli_channel(
+                mover_px, mover_py, mover_pz, movers
+            )
+
+            def pair_qubit(i: int):
+                return ilist.IList([movers[i], sitters[i]])
+
+            groups = ilist.map(pair_qubit, ilist.range(len(movers)))
```
jasonhan3: I tried to test this and due to some compile-time artifact, kirin infers `len(movers)` to be 0, so groups is an empty list. I think to fix it you can change this to

```
 def pair_qubit(i: int):
-    return ilist.IList([movers[i], sitters[i]])
+    return ilist.IList([movers[i], cz_sitters[i]])

-groups = ilist.map(pair_qubit, ilist.range(len(movers)))
+groups = ilist.map(
+    pair_qubit,
+    ilist.range(len(mover_ids) * num_rows),
+)
```

jasonhan3 (changes requested):
Mainly a bug with len(movers) having length 0 for some reason

python/bloqade/lanes/arch/gemini/logical/upstream.py
```diff
@@ -313,7 +321,7 @@ def gate_with_noise(indices: ilist.IList[int, Any], gate):
     AggressiveUnroll(noisy_initialize_broadcast.dialects).fixpoint(
         noisy_initialize_broadcast
     )
-    # noisy_initialize_broadcast.print()
+    noisy_initialize_broadcast.print()
```
jasonhan3: nit: maybe comment out this print statement?

jasonhan3 (approved):
LGTM!

PR #820, opened 2026-07-21
Additional changes to refactoring simulator API

jasonhan3:
Additional changes on top of https://github.com/QuEraComputing/bloqade-lanes/pull/790 for refactoring simulator API

This PR deliberately narrows the Gemini simulator API around backend-owned execution: logical and physical simulators remain separate compilation paths, but their tasks now capture only the sampling backend rather than the whole simulator, and `task()` accepts only caller-prepared SQuin kernels—CUDA-Q conversion and any detector/observable annotation are explicitly outside this API. Sampling has one common result protocol, while Tsim’s detector-only path is configured on `TsimSimulatorBackend` rather than changing the per-run API; CliffT and PyQrack continue to return measurement-backed results. Per-call seeds are preferred so `run`, `run_async`, and backend sampling can be deterministic without carrying public mutable seed configuration. The main deferred architectural issue remains the same as in [[#790’s design summary](https://github.com/QuEraComputing/bloqade-lanes/pull/790)](https://github.com/QuEraComputing/bloqade-lanes/pull/790): CliffT and PyQrack still compose Tsim for SQuin translation and detector-error-model generation, so a future refactor should separate translation, sampling, and DEM-generation capabilities rather than making non-Tsim samplers depend on Tsim.


jasonhan3:
Summary of commits on top of base branch:
- manual merging of 'main' into this branch; basically just added files and did not resolve merge conflicts. Doesn't really need thorough review.
- resolved merge conflict by manually adding "seed" argument to task.run(). This was a feature added in 'main' which I also had to add to pyqracksimulator. Does need a bit more careful review as I'm basically adding the feature of "adding a seed to run() methods".
- resolved merge conflict manually added imports for some of the shared helpers defined in physical_simulator.py. doesn't really need review.
- breaking change to remove support for m2dets and m2obs on task() on geminilogicalsimulator. by not supporting cuda-q kernels on the logical simulator we standardize the API so we don't have to support it for the physical simulator either. We could in principle support cuda-q conversion here, but after talking with stefan and phil it didn't seem like a good idea. We could have alternatively defined a "task_cudaq" method, but I decided against it because this would imply that we would implement additional conversions from other kinds of kernels by defining a new "task_<program_type>", which is probably not the API we want to expose. the design decision needs review but the implementation not so much.
- breaking change to make tsim_backend private in simulator backends; we ideally do not want every simulator backend to need to be composed of tsim_backend for the future. doesn't really need review.
- breaking change to make detector_error_model method private on the simulator backend. it doesn't really make sense for a simulator backend to have a _detector_error_model method. doesn't really need review.
- planning commit for removing run_detectors from public API's; doesn't need review
- another planning commit for removing run_detectors from public API's; doesn't need review
- commit to remove run_detectors from public API and make it a field in tsimbackend. the idea is that I don't think that run_detectors should be part of the public API across simulator backends, simulator tasks, and simulators because it really is a tsim-specific configuration, to my understanding. thus, I've tried to localize the "run_detectors" option to tsim by making it a field in tsim backend. **this is probably the most substantial design change in this series of commits and so would require review**
- breaking change to make simulator tasks store the simulation backend instead of the simulator as the simulation backend appears to be all they need; doesn't require careful review
- breaking change to make _seed a private field in clifftbackend and pyqrackbackend because we don't want to support seeds on the simulators (this doesn't really make sense) but rather want to focus on supporting seeds when you call sample() or run(). doesn't really need review
- commit to update the msd_postselection_experiment.py file in the "demo" folder to account for the private seed and instead use the seed API on .run(). doesn't really need review; forgot to include in the previous commit
- commit to add back docstrings that were deleted. doesn't really need review
- commit to delete plans/specs for implemented commits; doesn't really need review
- commit to change test for pyqrack; getting some unexpected behavior with the seeds. 

TODO for the future
- correctly implement pyqrack "seed" functionality.

weinbe58 (approved):
Thanks!

PR #824, opened 2026-07-22
fix(python): index terminal measurements by record ID, not qubit ID

### LINKED ISSUE #815: Atom postprocessing indexes raw measurements by qubit ID instead of measurement-record ID
# Atom postprocessing indexes raw measurements by qubit ID instead of measurement-record ID

## Summary

`PhysicalSimulator` and `GeminiLogicalSimulator` can produce incorrect return
values, detectors, and observables when qubits are measured in an order different
from their allocation order.

Simulator backends return raw measurements in measurement-execution order, but
Atom postprocessing indexes those results using `qubit_id`. These indices only
coincide when qubits are measured once in allocation order.

## Reproduction

```python
from bloqade import qubit, squin
from bloqade.gemini import logical
from bloqade.gemini.device import (
    GeminiLogicalSimulator,
    PhysicalSimulator,
)


@squin.kernel
def physical_kernel():
    q = squin.qalloc(3)
    squin.x(q[2])

    # Measurement-record order: q2, q0, q1
    m = squin.broadcast.measure(ilist.IList([q[2], q[0], q[1]]))

    # m[0] is q2 and should be True.
    squin.set_detector(ilist.IList([m[0]]), [0])
    squin.set_observable(ilist.IList([m[0]]))

    return m


@logical.kernel(aggressive_unroll=True)
def logical_kernel():
    q = qubit.qalloc(3)
    squin.x(q[2])

    # Logical measurement order: q2, q0, q1
    m = logical.terminal_measure(ilist.IList([q[2], q[0], q[1]]))

    # Logical-Z parity for m[0], i.e. logical q2.
    logical_z = ilist.IList([m[0][0], m[0][1], m[0][5]])

    squin.set_detector(logical_z, [0])
    squin.set_observable(logical_z)

    return m


def compare(label, task):
    postprocessed = task.run(
        shots=3,
        with_noise=False,
        run_detectors=False,
    )

    native = task.run(
        shots=3,
        with_noise=False,
        run_detectors=True,
    )

    print(f"\n{label}")
    print("postprocessed detectors: ", postprocessed.detectors)
    print("postprocessed observables:", postprocessed.observables)
    print("native detectors:        ", native.detectors)
    print("native observables:      ", native.observables)


compare(
    "PhysicalSimulator",
    PhysicalSimulator().task(physical_kernel),
)

compare(
    "GeminiLogicalSimulator",
    GeminiLogicalSimulator().task(logical_kernel),
)
```

## Actual behavior

```text
PhysicalSimulator
postprocessed detectors:  [[False], [False], [False]]
postprocessed observables: [[False], [False], [False]]
native detectors:         ((True,), (True,), (True,))
native observables:       ((True,), (True,), (True,))

GeminiLogicalSimulator
postprocessed detectors:  [[False], [False], [False]]
postprocessed observables: [[False], [False], [False]]
native detectors:         ((True,), (True,), (True,))
native observables:       ((True,), (True,), (True,))
```

## Expected behavior

The measurement-result contract is:

```text
m[0] = measurement of q2 = raw[0]
m[1] = measurement of q0 = raw[1]
m[2] = measurement of q1 = raw[2]
```

Therefore, detectors and observables referencing `m[0]` should be `True`.
The `run_detectors=False` results should agree with Tsim's native
`run_detectors=True` results.

## Root cause

Atom analysis represents a measurement result as:

```python
@dataclass
class MeasureResult:
    qubit_id: int
    location_address: LocationAddress
```

When analyzing `move.GetFutureResult`, it records which atom occupied the measured
location:

```python
MeasureResult(qubit_id, location_address)
```

The global measurement-record index is not preserved.

Atom postprocessing later generates:

```python
def _get_measurement(measurements):
    return measurements[elem.qubit_id]
```

This incorrectly treats the measured atom's ID as the raw measurement column.

For the reproduction above:

```text
raw[0] = q2
raw[1] = q0
raw[2] = q1
```

but the result associated with `m[0]` is represented as
`MeasureResult(qubit_id=2, ...)`, causing postprocessing to read `raw[2]`
instead of `raw[0]`.

## Why this is supported behavior

Non-allocation-order terminal measurement is intentionally supported. Existing
regression coverage requires physical compilation to preserve reversed and
arbitrary measurement permutations.

The Gemini restriction is that all allocated qubits must be measured in one
terminal measurement. It does not require allocation order.

Annotations reference measurement results, not qubit IDs:

```python
m = squin.broadcast.measure([q2, q0, q1])
squin.set_detector([m[0], m[1]], ...)
```

This must mean:

```text
measurement(q2) XOR measurement(q0)
```

Users should not need to compensate for the measurement permutation.

## Affected paths

- `PhysicalSimulator(...).run(run_detectors=False)`:
  - `return_values`
  - `detectors`
  - `observables`
- `GeminiLogicalSimulator(...).run(run_detectors=False)`:
  - `return_values`
  - `detectors`
  - `observables`

## Unaffected paths

- `result.measurements`; raw backend measurement ordering is correct.
- Tsim with `run_detectors=True`; Tsim returns native detector and observable
  arrays and bypasses Atom postprocessing.

## Suggested fix

Preserve a global measurement-record ID on each symbolic measurement result:

```python
@dataclass
class MeasureResult:
    measurement_id: int
    qubit_id: int
    location_address: LocationAddress
```

The ID must follow the order in which physical measurement operations execute:

```text
first measurement  -> measurement_id 0
second measurement -> measurement_id 1
...
```

Postprocessing should then use:

```python
measurements[elem.measurement_id]
```

rather than:

```python
measurements[elem.qubit_id]
```

`bloqade.gemini.post_processing.generate_post_processing` already demonstrates
this model with `RawMeasureId.idx`, although the current simulator tasks use
`AtomInterpreter.get_post_processing`.

## Acceptance criteria

- `PhysicalSimulator` correctly reconstructs return values for arbitrary terminal
  measurement permutations.
- `GeminiLogicalSimulator` correctly reconstructs logical return values for
  arbitrary terminal measurement permutations.
- Detectors and observables derived from raw measurements match Tsim's native
  `run_detectors=True` results.
- Existing allocation-order behavior remains unchanged.
- Regression tests cover:
  - allocation order;
  - reversed order;
  - a nontrivial arbitrary permutation;
  - both physical and logical simulator paths.

weinbe58:
## Summary

Fixes #815. `PhysicalSimulator` and `GeminiLogicalSimulator` produced incorrect return values, detectors, and observables whenever qubits were measured in an order different from allocation order.

Simulator backends return raw measurements in **measurement-execution order**, but Atom post-processing indexed those results by `qubit_id`. Those indices coincide only when qubits are measured once in allocation order.

## Root cause

The move-level IR already preserves the measurement permutation faithfully — the order of `move.GetFutureResult` statements (and the return `IList` that gathers them) matches the user's measurement order. But that ordinal was only *implicit* in statement order; nothing carried it into post-processing, which reconstructed a column index from `qubit_id` (a property of *which atom sat at the location*), discarding the order the IR preserved.

## Fix

- Add `measurement_id` to `MeasureResult`, assigned by a monotonic counter in `get_future_result_impl` as the analysis walks `GetFutureResult` in IR order.
- This stays in lockstep with the `qubit.measure` statements `InsertMeasurements` emits (same statements, same walk order), so `measurement_id` equals the raw measurement column index.
- Post-processing now uses `measurements[measurement_id]` instead of `measurements[qubit_id]`.

`InsertMeasurements` still uses `qubit_id` (it selects the physical qubit SSA), and `_shot_remapping` is untouched (it projects on `location_address`).

## Testing

New regression test `python/tests/gemini/test_measurement_permutation_regression.py` covers **allocation / reversed / arbitrary** measurement orders across **both** the physical and logical simulator paths, asserting post-processed detectors/observables match Tsim's native `run_detectors=True` results and that return values match the hand-computed truth.

- `analysis` + `rewrite` + `gemini` suites: 523 passed, 8 skipped.
- black / isort / ruff / pyright clean.

## Acceptance criteria

- [x] `PhysicalSimulator` correctly reconstructs return values for arbitrary terminal measurement permutations.
- [x] `GeminiLogicalSimulator` correctly reconstructs logical return values for arbitrary permutations.
- [x] Detectors and observables derived from raw measurements match Tsim's native `run_detectors=True` results.
- [x] Existing allocation-order behavior remains unchanged.
- [x] Regression tests cover allocation / reversed / arbitrary order, both physical and logical paths.


python/bloqade/lanes/analysis/atom/_post_processing.py
weinbe58, line 24: Can we toggle which version to index by because it will change based on the context. 

```suggestion
def constructor_function(
elem: MoveExecution,
*
use_qubit_id: bool = False
) -> Callable[[Sequence[bool]], Any] | None:
```

jasonhan3: @weinbe58 I made a pretty trivial change to allow indexing by the `qubit_id`, but I'm not sure if this is what we wanted? Because `_get_measurement` only takes in a sequence of booleans (which doesn't have a notion of `measurement_id` or `qubit_id` anymore), is the idea that `measurements` might be sorted in a different way such that using `qubit_id` is preferable?

jasonhan3 (approved):
LGTM!

PR #826, opened 2026-07-22
Jasonh/add plotting physical arch

jasonhan3:
Added plotting code for buses to "demo/physical_arch_customization.py"


weinbe58 (approved):
LGTM. 

We shuold think about formalizing the ways of visualizing the architecture spcification.

PR #831, opened 2026-07-23
Add demo notebooks to CI as well as integration test

jasonhan3:
- Add some demo notebooks to CI
- Add integration test for PostselectionExperiment

**NOTE: the integration test for PostselectionExperiment is kind of like the "benchmarks": if the compiler changes, or the noise model changes, that will implicitly affect this integration test, and so the test will need to be updated.**


weinbe58 (approved):
LGTM, I think as a follow up issue I should go through and fix the just file to make sure we're using `uv run --locked python` but this should not be a blocker here.

PR #838, opened 2026-07-28
feat: added integration for PPVM simulator backend

jasonhan3:
Notes:
- ~~Is a draft for now because uv.lock had a lot of changes due to a bug where `pyproject.toml` in PPVM installs `kirin` instead of `kirin-toolchain`, causing a bunch of extra unnecessary dependencies to be installed~~ -- Resolved with recent PPVM PR https://github.com/QuEraComputing/ppvm/pull/199
- Adds `PPVMSimulatorBackend` as another `AbstractSimulatorBackend`
- `PPVMSimulatorBackend`'s `sample` returns `measurements` with 3 values: False (|0>), True (|1>), and None (atom loss).
- Modifies `_normalize_matrix` in `GeminiPhysicalSimulator` and `GeminiLogicalSimulator` to add postprocessing to convert the `None` values to `True` to model how Gemini-MVP doesn't support loss-resolved readout.


weinbe58:
@jasonhan3 can you use the conventional commit convention when creating the title for the PR

python/bloqade/gemini/device/simulator.py
```diff
@@ -78,6 +80,38 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
             add_noise=False,
         ).emit(self.physical_move_kernel)
 
+    @staticmethod
+    def _normalize_matrix(
+        payload: Any,
+        *,
+        name: str,
+        shots: int,
+    ) -> list[list[bool]]:
```
weinbe58, line 90: Can you rename this function or add a short doc string describing what this is for? it isn't obvious at first glance what the loss value is in the output results nor what value loss appears in the payload. It is also not clear what value  you're converting the loss value to.

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -190,6 +192,38 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
             arch_spec=self.physical_arch_spec,
         ).emit(self.physical_move_kernel)
 
+    @staticmethod
+    def _normalize_matrix(
+        payload: Any,
+        *,
+        name: str,
+        shots: int,
+    ) -> list[list[bool]]:
```
weinbe58, line 202: This looks like duplicate code from the other class. should this be a helper function instead. Also maybe add fields `loss = None` and `loss_replace = True` instead.

python/bloqade/gemini/device/physical_simulator.py
```diff
@@ -190,6 +192,38 @@ def noiseless_physical_squin_kernel(self) -> ir.Method[[], RetType]:
...
+            array = np.asarray(payload, dtype=object)
+
+            # NOTE: this is to model a lack of loss-resolved readout; atom loss is indistinguishable from measuring the |1> state
+            none_mask = np.fromiter(
+                (value is None for value in array.flat),
+                dtype=bool,
+                count=array.size,
+            ).reshape(array.shape)
+
+            array[none_mask] = True
```
weinbe58:
```suggestion
            array[array == None] == True
            array = array.astype(bool)
```

this should also work

python/bloqade/gemini/device/simulator_backend.py
```diff
@@ -191,6 +191,43 @@ def _clifft_compatible_stim_text(circuit: Any) -> str:
...
+_PPVM_CORRELATED_LOSS_EVENT = re.compile(
+    r"^(?P<instruction>\s*I_ERROR)"
+    r"\[correlated_loss:(?P<event_id>\d+)\]"
+    r"(?P<arguments>\([^)\n]*\))"
+    r"(?P<targets>\s+[^#\n]+)"
+    r"(?P<comment>#.*)?$"
+)
+
+
+def _ppvm_compatible_stim_text(circuit: Any) -> str:
```
weinbe58: Can you add expected types for the input here?

weinbe58 (changes requested):
Needs a little bit of work but overall looks good!

python/bloqade/gemini/device/physical_simulator.py
jasonhan3: I think there are some small edge cases where it's safer to check that the `value is None` rather than `value == None`, so I'll stick to the existing implementation for now

weinbe58 (approved):
Approving this but we will h

PR #852, opened 2026-07-30
perf(search): speed up entropy driver candidate generation 2.5-4.7x

weinbe58:
## Summary

The entropy-guided move driver (`crates/bloqade-lanes-search/src/drivers/entropy.rs`) spent nearly all of its wall time in candidate generation. Profiling with a new micro-benchmark surfaced two sources of redundant work; both fixes are **behavior-preserving** (identical search results).

### 1. Drop the redundant `generate_candidates` call in `get_next_candidate`
It regenerated the candidate cache **twice** per call. `generate_candidates` is a pure function of `(config, entropy, params, ctx, seed)` — none of which change between the two calls — so the second call produced an identical cache with the same all-tried outcome. Now it regenerates once, and only regenerates again when the cache just scanned was stale (carried over from a lower entropy). → roughly halves candidate-generation calls.

### 2. Cache `BusGridContext`'s arch maps in `LaneIndex`
The four lookup maps (`pos_to_src`, `src_to_lane`, `src_to_dst`, `src_to_pos`) depend only on `(index, move_type, bus_id, direction)` — **not** on occupancy — yet `BusGridContext::new` rebuilt them (an all-lanes scan) on every call: ~5–8× per `generate_candidates`, thousands of times per solve. They're now precomputed once per bus group in `LaneIndex` (`BusGridMaps`) and borrowed via `Cow`. The constructor signature is unchanged, so all five call sites are untouched.

## Results

Driver-only wall time on the new micro-benchmark (`full.json` arch), before → after:

| scenario | baseline | optimized | speedup |
|---|---:|---:|---:|
| single_multistep | 108,803 µs | 24,047 µs | **4.5×** |
| route4_parallel | 387,929 µs | 118,739 µs | **3.3×** |
| route4_intraword | 385,871 µs | 124,653 µs | **3.1×** |
| route6_cross | 453,672 µs | 183,094 µs | **2.5×** |
| route8_dense | 919,144 µs | 367,868 µs | **2.5×** |

A perf-neutral attempt (flattening `DistanceTable::time_distance`) was measured and reverted.

## Verification (behavior-preserving)

- **Identical fingerprint** (`nodes_expanded` / goal depth) across every benchmark scenario, before vs. after.
- **Rust suite**: 587 passed, 0 failed (core, cli, dsl-core, search).
- **Python suite**: 1813 passed, 0 failed (against the rebuilt extension).
- **Routing benchmarks**: both `just benchmark-physical` and `just benchmark-logical` report **"no differences found"** — every deterministic metric (`success`, `move_count_events`, `move_count_lanes`, `estimated_fidelity`, `nodes_explored`, `max_depth_reached`) is unchanged, so the committed baselines need no regeneration.

## Notes

- Adds a self-contained micro-benchmark `crates/bloqade-lanes-search/benches/entropy.rs` (`harness = false`, no new dependencies) that prints the search fingerprint as a regression guard: `cargo bench -p bloqade-lanes-search --bench entropy`.
- Follow-up (not in this PR, since it would be behavior-changing): after finding its first goal, the driver spins to the hard iteration cap trying to reach `max_goal_candidates` — a 2-move solve runs the full 4000 iterations. Eliminating that spin is the largest remaining lever but needs a provably result-preserving termination condition plus baseline regeneration.

### DIFF
--- a/crates/bloqade-lanes-search/src/primitives/lane_index.rs
+++ b/crates/bloqade-lanes-search/src/primitives/lane_index.rs
@@ -9,7 +9,7 @@ use std::collections::HashMap;
 use bloqade_lanes_bytecode_core::arch::addr::{Direction, LaneAddr, LocationAddr, MoveType};
 use bloqade_lanes_bytecode_core::arch::types::ArchSpec;
 
-use crate::ops::aod_grid::BusGridMaps;
+use crate::primitives::bus_grid_maps::BusGridMaps;
 
 /// Precomputed lane lookups for an architecture.
 ///
--- a/crates/bloqade-lanes-search/src/primitives/lane_index.rs
+++ b/crates/bloqade-lanes-search/src/primitives/lane_index.rs
@@ -9,6 +9,8 @@ use std::collections::HashMap;
 use bloqade_lanes_bytecode_core::arch::addr::{Direction, LaneAddr, LocationAddr, MoveType};
 use bloqade_lanes_bytecode_core::arch::types::ArchSpec;
 
+use crate::ops::aod_grid::BusGridMaps;
+
 /// Precomputed lane lookups for an architecture.
 ///
 /// Built once from an [`ArchSpec`] and reused across multiple searches.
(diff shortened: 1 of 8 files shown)


crates/bloqade-lanes-search/src/primitives/lane_index.rs
```diff
@@ -34,6 +36,11 @@ pub struct LaneIndex {
     lane_durations: HashMap<u64, f64>,
     /// Fastest lane duration across all lanes with paths. `None` if no paths.
     fastest_lane_duration: Option<f64>,
+    /// Precomputed AOD-grid lookup maps per `(move_type, bus_id, direction)`
+    /// bus group, spanning all zones. Occupancy-independent, so they are
+    /// built once here and borrowed by every `BusGridContext` for that group
+    /// (see [`BusGridMaps`]).
+    bus_grid_maps: HashMap<(MoveType, u32, Direction), BusGridMaps>,
```
weinbe58: Can we use TripletKey type for the key?

PR #865, opened 2026-08-03
feat: added qalloc_at stdlib kernel

### LINKED ISSUE #854: Have a way to allocate logical qubits by (zone, row, col)
The `loc` statement is currently supported for physical kernels/architecture specifications, but is not currently supported for logical kernels.

In principle it works:

``` python

from bloqade.gemini import logical as gemini_logical
from bloqade.gemini.common.dialects.qubit import new_at
from bloqade.gemini.device import GeminiLogicalSimulator
from bloqade.lanes.dialects import arch

# A local logical-kernel dialect group that additionally permits loc(...).
logical_with_loc = gemini_logical.kernel.union([arch])


@logical_with_loc
def coordinate_pinned_logical_qubit():
    # (zone, row, col); row can be between 0-4 and col can be only be 0 or 2.
    address = arch.loc(0, 1, 2)

    # loc resolves to logical-architecture (zone, word, site) at compile time.
    q = new_at(address.zone_id, address.word_id, address.site_id)

    # Logical kernels use the logical terminal-measure operation.
    gemini_logical.terminal_measure(ilist.IList([q]))


task = GeminiLogicalSimulator().task(coordinate_pinned_logical_qubit)

# Compilation succeeded; inspect the compiled movement program if useful.
task.physical_move_kernel
```

but it assumes that all qubits are allocated in a row, and that we are specifying the location of the leftmost logical qubit in the Steane code.

Two points:
1. Can add this `loc` statement to the logical kernel dialect to support `loc`
2. Potentially a larger discussion for how to specify allocating logical qubits in the kernel

jasonhan3:
closes https://github.com/QuEraComputing/bloqade-lanes/issues/854

- Adds "qalloc_at" standard library kernel for the logical kernel to reduce syntax needed to allocate logical qubits at specific positions.

### DIFF
--- a/python/bloqade/gemini/logical/stdlib/qubit_allocation.py
+++ b/python/bloqade/gemini/logical/stdlib/qubit_allocation.py
@@ -1,22 +1,23 @@
-from typing import TypeVar
+from typing import Optional, TypeAlias, TypeVar
 
 from bloqade.types import Qubit
-from kirin import types
 from kirin.dialects import ilist
 
-from bloqade import squin
+from bloqade import qubit
 from bloqade.gemini.common.dialects.qubit import new_at
 
 from .. import kernel
 
 N = TypeVar("N")
 
-PositionIndex: Alias = int | None
+# Kirin preserves the static ``IList`` length through ``ilist.map`` for
+# ``typing.Optional`` but not the equivalent PEP 604 ``int | None`` spelling.
+PositionIndex: TypeAlias = Optional[int]  # noqa: UP045
 
 
 @kernel(aggressive_unroll=True, verify=False)
 def qalloc_at(
-    positions: ilist.IList[PositionIndex, N],  # pyright: ignore[reportInvalidTypeForm]
+    positions: ilist.IList[PositionIndex, N],
 ) -> ilist.IList[Qubit, N]:
     """Allocate logical qubits at optional linear logical positions.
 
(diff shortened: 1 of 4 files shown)


python/bloqade/gemini/logical/stdlib/qubit_allocation.py
```diff
@@ -0,0 +1,39 @@
...
+    """
+
+    def position_to_new_at(
+        position: PositionIndex,  # pyright: ignore[reportInvalidTypeForm]
+    ) -> Qubit:
+        if position is None:
+            qubit = squin.qalloc(1)[0]
+        else:
+            qubit = new_at(0, position * 2, 0)
+        return qubit
```
weinbe58:
```suggestion
        if position is None:
            q = qubit.new()
        else:
            q = new_at(0, position * 2, 0)
        return q
```

you can use the `bloqade.qubit` dialect to allocate a single qubit so we don't need to fold a whole subroutine call extra here.

python/bloqade/gemini/logical/stdlib/qubit_allocation.py
```diff
@@ -0,0 +1,39 @@
...
+from kirin.dialects import ilist
+
+from bloqade import squin
+from bloqade.gemini.common.dialects.qubit import new_at
+
+from .. import kernel
+
+N = TypeVar("N")
+
+PositionIndex = types.Union(types.Int, types.NoneType)
```
weinbe58: Do not use kirin types here this is not valid. 
```suggestion
PositionIndex: Alias = int | None
```

weinbe58 (changes requested):
Need some small fixes

jasonhan3:
Hi @weinbe58  -- made the changes! Needed to make a small change to use `Optional[int]` due to some issues with Kirin (https://github.com/QuEraComputing/kirin/issues/692)

weinbe58 (approved):
LGTM

PR #869, opened 2026-08-04
feat(search): add Push and Rotate as a complete router and opt-in fallback

weinbe58:
## Summary

Adds a **Push and Rotate** router (de Wilde, ter Mors & Witteveen, JAIR 51, 2014) to `bloqade-lanes-search`: a rule-based, complete multi-agent pathfinding planner that composes with the existing solver surface, plus wiring that puts it to work in three places:

- **A standalone strategy** (`Strategy::PushRotate` / `solve_push_rotate`) — fixed-target routing, ~two orders of magnitude faster than the search strategies, at the cost of more AOD operations.
- **An opt-in reliability net** (`SolveOptions::fallback_push_rotate`, off by default): when a search strategy fails, the planner converts "gave up" into either a valid schedule or — within its regime — a *proof* that none exists.
- **The entropy driver's budget-exhaustion fallback**: the old greedy sequential router (which could not displace atoms out of each other's way) now runs Push and Rotate first, keeping the greedy router only for its one residual case (a direct slide into a single hole at `m = 1`, outside the planner's regime).

The planner builds on the Kornhauser decomposition shared with the feasibility oracle (#867), so the two can never disagree about an instance's structure.

## Verdict semantics

`SolveStatus::Unsolvable` from this router is a **proof**, and the fallback path promotes it over the search's verdict on that basis — so only genuine proofs may produce it:

| verdict | meaning |
|---|---|
| `Unsolvable` | proven: subgraph confinement (both time directions), no path, or a target for a nonexistent qubit |
| `BudgetExceeded` | not proven: out of the ≥ 2-empties-per-component regime, planner stuck, or scheduling artifact |

Two documented caveats: proofs are **relative to the blocked set** (spectators-as-blocked reads as "unsolvable without touching spectators"), and the planner does not prove *every* unsolvable in-regime instance — the paper's `f = f'` test is unsound under the deliberately conservative shared `assign_agents`, so the containment form is used; closing that gap needs a planner-specific paper-faithful assignment (follow-up).

## Invalid requests rejected at the boundary

Non-injective requests (two qubits sharing a target location, one qubit with two targets, two qubits on one initial location) are rejected as `ConfigError` at the solve entry points, before any feasibility, search, or planning pass — they are caller errors, never verdicts. The feasibility module's obstructions remain as defence in depth.

## Testing

- **End-to-end replay**: plans validated through `AtomStateData` (the same simulator the IR pipeline uses) — per-operation `check_lanes`, collision-freedom, lanes-issued == atom-moves, exact final placement — over walk instances, permutations, and both shipped Gemini specs, plus blocked-spectator instances and the packaged `move_layers` from the solver surface.
- **Brute-force property test**: 120 carved instances (≤ 8 vertices, blocked sets that often disconnect the graph) against a configuration-space BFS oracle: `Unsolvable` ⇒ oracle-unsolvable, `Solved` ⇒ replays to the goal, in-regime ∧ solvable ⇒ `Solved` (currently zero misses). This test caught and now pins three real bugs (unsound `f = f'`, over-restricted swap candidates, stale-`q` rotate aborts).
- Verdict-mapping, fallback-promotion, phantom-target, out-of-regime, and invalid-request tests at the solver surface.
- **Benchmarks**: both suites regenerated and compared — no deterministic differences; committed baselines unchanged (the router is opt-in and no benchmark case's outcome depended on the old fallback).

## Notes for review

- The design note (`docs/superpowers/specs/2026-08-02-push-rotate-router-design.md`) documents the five deviations from the paper with rationale, the heuristic seam ("rank, don't choose" — a bad heuristic is a quality problem, never a correctness one), and measured negative results.
- Deferred polish from pre-PR review: `GroupKey` dedup, `AlignmentHeuristics` cache-staleness guard, scheduler worklist.
- Non-breaking: new module + opt-in wiring; default solver behaviour unchanged.

### DIFF
--- a/crates/bloqade-lanes-search/src/push_rotate/heuristics.rs
+++ b/crates/bloqade-lanes-search/src/push_rotate/heuristics.rs
@@ -1,0 +1,279 @@
+//! Swappable decision points in the planner.
+//!
+//! Push and Rotate has no objective function — it is a completeness algorithm.
+//! §5 of de Wilde, ter Mors & Witteveen identifies the four places where a
+//! choice is free and affects solution quality, and this trait is exactly
+//! those four:
+//!
+//! | § | Decision | Method |
+//! |---|---|---|
+//! | 1 | which empty vertex `clear_vertex` drains into | [`PlanHeuristics::rank_clear_target`] |
+//! | 2 | which degree-≥3 vertex `swap` uses | [`PlanHeuristics::rank_swap_vertex`] |
+//! | 3 | the order agents are planned in | [`PlanHeuristics::agent_order`] |
+//! | 4 | which shortest path an agent takes | [`PlanHeuristics::score_step`] |
+//!
+//! The paper's authors explored only 3 and 4, and noted that agent ordering
+//! "can be very important to solution quality". [`DefaultHeuristics`] does
+//! neither — it is the unoptimised baseline the current benchmark numbers come
+//! from.
+//!
+//! ## Why scores rather than choices
+//!
+//! Every method returns a *ranking key* instead of making the choice, and the
+//! caller applies it as a tie-break on top of the existing rule. That keeps a
+//! heuristic from breaking correctness: it can reorder equally-good options
+//! but cannot select an illegal one or lengthen a path. A heuristic that
+//! returns a constant is exactly the default.
+//!
+//! ## The one that matters next
+//!
+//! [`PlanHeuristics::score_step`] is where AOD alignment goes. `headroom`
+//! shows the dependency DAG admits ~5x more parallelism than the scheduler
+//! extracts, and the gap is entirely geometric: moves that are ready together
+//! do not share a bus group or do not form a rectangle. Preferring steps that
+//! align with other pending moves is what closes it, and [`PlanCtx::edge`]
+//! exposes the bus group and source position needed to judge that.
+
+use std::cell::RefCell;
+use std::collections::HashMap;
+
+use crate::feasibility::graph::VertexId;
+
+use crate::push_rotate::context::{GroupKey, PlanCtx};
+use crate::push_rotate::state::PlanState;
+
+/// Strategy for the planner's free choices.
+///
+/// All methods have defaults reproducing the current behaviour, so an
+/// implementation need only override what it cares about.
+pub trait PlanHeuristics {
+    /// Order in which agents are planned.
+    ///
+    /// Must contain every agent exactly once. Subgraph precedence still
+    /// applies on top: agents assigned to no subgraph are planned last
+    /// regardless (§3.1.3).
+    fn agent_order(&self, ctx: &PlanCtx, state: &PlanState) -> Vec<u32> {
+        let _ = state;
+        let n = ctx.goal.len();
+        let mut assigned: Vec<u32> = Vec::new();
+        let mut unassigned: Vec<u32> = Vec::new();
+        for a in 0..n as u32 {
+            match ctx.decomp.assignment.get(&a) {
+                Some(_) => assigned.push(a),
+                None => unassigned.push(a),
+            }
+        }
+        assigned.sort_by_key(|a| {
+            (
+                ctx.decomp.assignment.get(a).copied().unwrap_or(usize::MAX),
+                *a,
+            )
+        });
+        assigned.extend(unassigned);
+        assigned
+    }
+
+    /// Preference for `agent` stepping `from -> to`, used to break ties
+    /// between equally short paths. **Higher is better.**
+    ///
+    /// Only consulted among steps that are already on a shortest path, so it
+    /// cannot lengthen a route.
+    fn score_step(
+        &self,
+        ctx: &PlanCtx,
+        state: &PlanState,
+        agent: u32,
+        from: VertexId,
+        to: VertexId,
+    ) -> f64 {
+        let _ = (ctx, state, agent, from, to);
+        0.0
+    }
+
+    /// Ranking key for an empty vertex that `clear_vertex` could drain into.
+    /// **Lower is better.** `hops` is its distance from the vertex being
+    /// cleared.
+    fn rank_clear_target(
+        &self,
+        ctx: &PlanCtx,
+        state: &PlanState,
+        clearing: VertexId,
+        candidate: VertexId,
+        hops: u32,
+    ) -> i64 {
+        let _ = (ctx, state, clearing, candidate);
+        hops as i64
+    }
+
+    /// Ranking key for a candidate swap vertex. **Lower is better.**
+    /// `hops` is its distance from the agent initiating the swap.
+    fn rank_swap_vertex(
+        &self,
+        ctx: &PlanCtx,
+        state: &PlanState,
+        r: u32,
+        s: u32,
+        candidate: VertexId,
+        hops: u32,
+    ) -> i64 {
+        let _ = (ctx, state, r, s, candidate);
+        hops as i64
+    }
+}
+
+/// The unoptimised baseline: nearest-first everywhere, arbitrary agent order.
+///
+/// Every method is the trait default, so this is the behaviour all current
+/// benchmark numbers were measured with. Keep it as the control when
+/// evaluating anything else.
+#[derive(Debug, Default, Clone, Copy)]
+pub struct DefaultHeuristics;
+
+impl PlanHeuristics for DefaultHeuristics {}
+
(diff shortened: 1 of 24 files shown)


crates/bloqade-lanes-search/src/push_rotate/solver.rs
```diff
@@ -0,0 +1,150 @@
...
+/// [`SolveStatus::BudgetExceeded`], which proves nothing.
+///
+/// The proof is **relative to `blocked`**: callers that encode spectator
+/// atoms as blocked locations (see `block_spectators` in the Python
+/// placement strategy) get "no solution that leaves the spectators
+/// untouched" — unblocking a spectator may make the instance solvable.
+///
+/// # Errors
+///
+/// Returns [`ConfigError`] if `initial` contains duplicate qubit IDs.
```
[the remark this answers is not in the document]
weinbe58: Fixed in 9b33a8a9 — the error contract now lists all four request-shape errors (duplicate qubit id, duplicate occupancy, duplicate target location, duplicate target qubit).

crates/bloqade-lanes-search/src/search/target_solver.rs
```diff
@@ -103,8 +107,27 @@ pub(crate) fn solve_with_engine(
...
+    // frontier/generator apparatus below rather than being a `Frontier`.
+    if opts.strategy == Strategy::PushRotate {
+        return solve_push_rotate(
+            engine.index(),
+            &initial_pairs,
+            &target_pairs,
+            &blocked_locs,
+            DEFAULT_MOVE_BUDGET,
+        );
+    }
```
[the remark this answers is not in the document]
weinbe58: Documented as deliberately ignored (9b33a8a9), in both the fast-path comment and `Strategy::PushRotate`'s public doc. Mapping was considered and rejected: `max_expansions` budgets search *node expansions*, while the planner's move budget is a runaway guard for a rule-based algorithm — reusing the (often small) expansion cap as a move cap would strangle exactly the solves this strategy and the fallback exist to finish (the fallback tests pass `Some(1)`).

crates/bloqade-lanes-search/src/push_rotate/schedule.rs
```diff
@@ -0,0 +1,295 @@
...
+/// One AOD operation: moves executed simultaneously, with the lanes realising
+/// them.
+#[derive(Debug, Clone)]
+pub struct Batch {
+    pub moves: Vec<Move>,
+    pub lanes: Vec<LaneAddr>,
+}
+
+/// Bus-group key. Matches `LaneIndex`'s own grouping.
+type GroupKey = (u8, u32, u32, u8);
```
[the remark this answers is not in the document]
weinbe58: Fixed in 9b33a8a9 — `schedule.rs` now reuses `context::GroupKey`.

crates/bloqade-lanes-search/src/push_rotate/heuristics.rs
```diff
@@ -0,0 +1,279 @@
...
+/// move, so it is computed once per solve. The per-bus support counts do
+/// change, and are rebuilt only when the placement advances —
+/// `PlanState::moves().len()` is a monotonic clock for that. A rebuild is
+/// `O(agents x degree)`, a few hundred operations, against the thousands of
+/// `score_step` calls a single BFS makes.
+pub struct AlignmentHeuristics {
+    /// Weight on routing onto a bus other atoms can also use.
+    pub bus_weight: f64,
+    /// Weight on preferring word buses, where most parallelism lives.
+    pub word_bus_weight: f64,
```
weinbe58: I'm not sure how words buses are more parallel? can you explain or validate that this is true

crates/bloqade-lanes-search/src/push_rotate/heuristics.rs
```diff
@@ -0,0 +1,279 @@
...
+
+        // Discount this agent's own contribution, so the score reflects how
+        // much *company* the step would have.
+        let peers = cache
+            .support
+            .get(&info.group)
+            .copied()
+            .unwrap_or(0)
+            .saturating_sub(1);
+        let _ = agent;
```
weinbe58: What is this here for?

crates/bloqade-lanes-search/src/push_rotate/schedule.rs
```diff
@@ -0,0 +1,293 @@
...
+    Some(out)
+}
+
+/// Largest subset of `cand` whose source positions form a complete X×Y grid.
+///
+/// Enumerates subsets of the distinct y values and, for each, intersects the
+/// x values present on those rows — the largest `|rows| × |common x|` wins.
+/// Exponential in the number of distinct rows, which is why it is capped;
+/// observed rectangles on Gemini top out at 3×3.
+fn largest_rectangle(cand: &[usize], src_pos: &[(u64, u64)]) -> Vec<usize> {
```
weinbe58, line 216: I'm wondering here if you can reuse the current aod_bus infrastructure to define the different rectangles instead of running your own.

crates/bloqade-lanes-search/src/push_rotate/smooth.rs
```diff
@@ -0,0 +1,331 @@
...
+//! ## Why this is safe
+//!
+//! Removing a round trip can only make later moves *more* legal, never less:
+//! a move's precondition is that its destination is empty, and deleting an
+//! agent's excursion only frees vertices earlier. The vertex the agent
+//! returned to is untouched by anyone else, which is exactly what the
+//! redundancy condition guarantees. Plans are replayed against the graph in
+//! the tests regardless.
+
+use std::collections::VecDeque;
```
weinbe58, line 68: One quetion I have is the order in which smooth and schedule are applied. its possible to have a situation where the heuristics insert more moves to improve parallelism but then the smooth pass remove those moves before the scheduler can effectively utilize those moves to improve parallel depth.

crates/bloqade-lanes-search/src/push_rotate/heuristics.rs
weinbe58: Good instinct to question it — I ablated the term on the design doc's own instances (physical and logical k16 × 5 seeds): `bus_weight` alone gives 159/240 total ops, adding `word_bus_weight=0.25` gives 159/240, default heuristics 163/257. **The word-bus preference contributed exactly nothing** — the entire improvement comes from the peer-support term. The claim was unvalidated intent, so the term is removed in c7b0bfdc and the ablation is recorded in the struct doc and the design note.

crates/bloqade-lanes-search/src/push_rotate/heuristics.rs
weinbe58: It suppressed the unused-parameter warning: `agent` is part of the `score_step` trait signature (a custom heuristic may score by agent identity), but this implementation scores only by bus-group company. Replaced with an `_agent` parameter name + a comment in c7b0bfdc.

crates/bloqade-lanes-search/src/push_rotate/schedule.rs
weinbe58: Feasible but not a drop-in — the two solve different constraint problems. `largest_rectangle` works on the *ready set* of the dependency DAG (occupancy validity already guaranteed), is exact within its 12-row cap, and emits only actual moves. `build_aod_grids` works on a *live occupancy*, is greedy+merge (approximate), and — the interesting part — **completes rectangles with empty filler lanes**, which could yield bigger batches than the scheduler can form today. Adopting it needs an occupancy snapshot per batch, breaks the tests' lanes==moves invariant as written, and may lose exactness on some ready sets, so it wants a measured experiment rather than an in-PR change. Filed as #870 with the full trade-off table.

crates/bloqade-lanes-search/src/push_rotate/smooth.rs
weinbe58: The specific scenario can't occur, by a documented invariant: heuristics are rank-only over shortest paths — `score_step` is consulted only among steps already on a shortest path, so a heuristic can reorder moves but never *insert* one. Everything smooth removes is a `clear`/swap round-trip side effect, not a parallelism investment. The reverse interaction is theoretically real (a removed leg could have been incidentally completing a rectangle, so deletion can cost more ops than it saves), but it has no surface while smoothing's measured contribution on Gemini is near zero (`Plan::raw_move_count` vs the smoothed count exists to keep that visible). Documented in smooth.rs (c7b0bfdc) along with the airtight fix if it ever changes: schedule both the smoothed and unsmoothed plans and keep the shallower.

Issue #874, opened 2026-08-05
`ArchSpec::validate()` accepts ill-formed buses (cyclic src→dst, duplicate src, duplicate dst)


weinbe58:
Part of #866 (Phase 2). Full audit table in [the Phase 1 comment](https://github.com/QuEraComputing/bloqade-lanes/issues/866#issuecomment-5196424415); target semantics in [the architecture comment](https://github.com/QuEraComputing/bloqade-lanes/issues/866#issuecomment-5196425263).

## Summary

A bus is by definition a set of explicit edge transports — never a permutation or rotation (AOD tones in one group share a direction and cannot cross). `ArchSpec::validate()` does not enforce any of the properties that make a bus well-defined:

1. **No acyclicity check.** A fully cyclic bus (`src=[0,1,2], dst=[1,2,0]`) — a spec that asks the hardware for a physically impossible rotation — validates cleanly. `check_zone_site_buses` / `check_zone_word_buses` walk `src` and `dst` in two separate, independent loops (`arch/validate.rs:246-253` vs `:254-261`; `:282-289` vs `:290-297`); the two sets are never brought into contact. The only joint src/dst predicate in the crate is the per-pair `zone_id` inequality for zone buses (`validate.rs:351-361`), which a 2-cycle across zones passes.
2. **Duplicate `src` entries are legal and silently shadow.** Endpoint resolution is first-linear-match (`arch/query.rs:127-132`), so a duplicated source hides later pairs. Acyclicity of *the* src→dst relation is not even well-defined without unique sources.
3. **Duplicate `dst` entries are legal.** Two lanes sharing a destination is a guaranteed collision with no physical counterpart.

Downstream group-level checks cannot compensate: `check_lane_group_geometry` (`arch/query.rs:859-902`) never resolves destinations (positions come from the lane's own fields; `lane_endpoints` is never called), and even a dst-side rectangle check would be vacuous on cycles, because a cycle's dst set is a *permutation of* its src set — cycles are invisible to any set-shape predicate and only detectable as a relation-level property.

## Evidence of intent never implemented

`ArchSpecError::GridInvariant` ("Grid invariant error (bus src/dst not rectangular)", `arch/validate.rs:30-32`) is declared and mapped to a Python exception name (`bytecode-python/src/errors.rs:29`) but **never constructed anywhere in the workspace** — the only place a joint src/dst invariant is even named.

## Consequences

[...]

Issue #866, opened 2026-08-03
Support AOD-compatible moves with overlapping bus src/dst, and make AtomStateData::apply_moves order-independent


weinbe58:
## Summary

Audit how overlapping bus endpoints (a bus whose destination set intersects its source set) are handled across every layer of the toolchain, surface the inconsistencies as concrete bug reports, then fix them as a series of focused PRs. The end goal is to **support AOD-compatible moves with overlapping endpoints** (conveyor-style chains) with consistent semantics everywhere, while keeping the move-feasibility oracle sound.

## Background

This came out of reviewing `feat/move-feasibility`, whose pebble-motion reduction rests on an architecture property: every bus maps a source set disjoint from its destination set (`validate_bus_disjointness`). Digging into whether that should become a format invariant revealed that the codebase is *already* split on the question — the AOD grid layer was written to accept overlapping groups, while other layers assume they cannot exist.

Two facts frame the whole discussion:

- **Chains vs. cycles.** A chain (`x→y, y→z`, final destination empty) is a physically natural AOD operation (shift a trapped row/column by one site) and is harmless to the feasibility oracle: it serializes into single-pebble moves, so it does not change reachability. A **cycle** (`x→y, y→z, z→x`, fully occupied) rotates atoms with no empty vertex — pebble motion can never do this, so cyclic move sets would falsify `Infeasible` verdicts. Crucially, a cyclic group is not a capability we are giving up: **the architecture already defines a bus as a set of explicit move operations along edges of the architecture graph — never a permutation or rotation** (physically, AOD tones in one group share a direction and cannot cross). Per-bus **acyclicity** of the src→dst relation is that existing semantic definition written as a checkable rule, not a new restriction. It needs to be executable rather than prose: once `apply_moves` gets simultaneous semantics, a cyclic pair set would otherwise execute as a rotation, whatever the docs say buses mean.
- **The shipped Gemini specs are disjoint**, so none of the overlap-handling code paths are reachable today. Every inconsistency below is latent until a custom or future spec uses overlapping buses.

## Phase 1 — Audit

Enumerate every layer that assumes something about endpoint overlap or move simultaneity, and record: what it assumes, what it does on an overlapping (acyclic) group, and what it does on a cyclic one. Known layers to cover:

- [ ] `ArchSpec::validate()` (`arch/validate.rs`) — currently silent on endpoint overlap entirely
[...]

PR #896, opened 2026-08-07
feat(search): assemble conveyor chains into a single AOD operation

### LINKED ISSUE #887: Make move generators assemble conveyor chains, not just tolerate them
Follow-up to #876 (part of #866). #876 made the search layers *agree* on the uniform destination rule and made conveyor chains **expressible**; it did not make every generator **assemble** them. This issue covers closing that gap.

## Current state after #876

| Layer | Chain support | Evidence |
|---|---|---|
| `BusGridContext::is_valid_rect` | Accepts a chain rectangle (mover-group exemption) | pre-existing |
| `ExhaustiveGenerator` | **Assembles chains** — enumerates every rectangle, so the 2-cell chain rect is emitted as one simultaneous move set | `generate_admits_conveyor_chain` |
| `HeuristicGenerator` | Admits chain lanes as *candidates* (the occupied-destination filter now defers to `occupant_can_vacate`), but **selection never co-selects the atom ahead**, so the chain is never assembled | `heuristic_selection_does_not_yet_assemble_chains` |
| Entropy driver | Filters only `blocked`, so chain lanes reach `is_valid_rect` — but whether its group formation actually assembles chains is **untested** | — |
| `GreedyGenerator` | Cannot express chains: `find_path_occupied` treats every atom as a wall, so the atom ahead is an obstacle rather than a follower | — |

The `HeuristicGenerator` case is concrete and reproducible: with a conveyor bus `0→1, 1→2, …`, atoms on sites 0 and 1, and targets 1 and 2, the generator emits only the follower's move (`1→2`). The leader waits for the next search step, so routing is still correct — it just takes two AOD operations where one would do.

## Why selection doesn't assemble chains

Movers are scored and selected independently per qubit. Nothing expresses "to move into an occupied site, the occupant must also be selected in this same group". `is_valid_rect` then rejects any rectangle where the leader was selected but the follower was not, so a chain only forms by coincidence.

## Scope

- [ ] Make heuristic selection chain-aware: when a candidate's destination is occupied by an atom that can vacate on the same bus group, co-select that atom's lane (transitively, to the end of the chain) or drop the candidate. This is a joint-selection problem, not a filter tweak.
- [ ] Decide whether `GreedyGenerator` should follow chains at all — it would need multi-agent pathfinding rather than single-atom BFS. Documenting it as chain-incapable is a legitimate outcome.
- [ ] Add entropy-driver coverage on an overlapping-bus spec: does its group formation assemble chains, and does it prefer them when they are strictly better?
- [ ] End-to-end routing test on a spec where a chain is the *only* way to make progress (e.g. a fully-packed corridor whose sole free site is at the head of the chain). Today the heuristic path would deadlock there while exhaustive would solve it — worth confirming and fixing.
- [ ] Benchmark a conveyor-capable synthetic spec to quantify what chain assembly buys (fewer AOD operations per episode).

## Non-goals

Cycle safety is already handled upstream: `ArchSpec::validate` rejects cyclic buses (#874), so generators only ever see chains. `is_valid_rect` itself has no acyclicity condition and does not need one for that reason.

## Fixtures already in place

`crate::test_utils::chain_arch_json()` (search crate) is a validated overlapping-bus spec — the example arch with site bus 0 rewired as `0→1, 1→2, 2→3, 3→4`. The feasibility module has a matching fixture plus random-walk and brute-force soundness tests on it.

weinbe58:
Closes #887.

Two related changes: rectangle growth learns to assemble conveyor chains, and
the arch spec is validated wherever it enters the search — the invariant that
makes chain assembly safe to do.

## 1. Chain assembly (#887)

Rectangle growth in `build_aod_grids` assumed `is_valid_rect` was monotone:
`greedy_init` added one entry at a time and set aside anything whose
*intermediate* rectangle failed. A conveyor chain's prefix is exactly what
fails — the leader's cell alone is invalid because the follower is not yet one
of the rectangle's sources — so the valid complete rectangle was never tried.
Every consumer routing through `build_aod_grids` (heuristic generator, entropy
driver, DSL pipeline) therefore serialised chains into one AOD operation per
hop, even though #866/#876 had already made the execution model accept them.

`rect_outcome` replaces the boolean check with a single-pass
`Valid` / `Repairable` / `Invalid` result that records which movers' cells
would supply a missing exemption. `try_add_point` grows the rectangle, pulls
those cells in, and re-validates to a fixpoint, since each repair forces a new
row or column whose own cells may need repairing in turn.

### Cost

`try_add_point` tracks the requested point in a `bool` and undoes exactly the
coordinates it inserted, so the allocation-free rollback #882 tuned for is
preserved on the common path. The repair buffers allocate only when a chain
actually needs repairing, which cannot happen on an endpoint-disjoint bus.

Measured on a synthetic all-chain row (debug build): repair costs 1.1–2.3x the
disjoint baseline with no widening trend, because a chain assembles in one
`try_add_point` call and later entries hit the already-covered skip. The
pre-existing grow-and-revalidate cost is itself quadratic in rectangle size;
this adds a bounded constant, not an order of magnitude.

## 2. Spec validation at every entry to the search

The search layers treat per-bus acyclicity and endpoint uniqueness (#874) as
*given* rather than checking them — and chain assembly leans on exactly that.
Rectangle growth exempts an occupied destination whose occupant moves in the
same shot, which is what makes a chain legal; on a cyclic bus it is also what
would make a physically impossible rotation look legal. So the invariant has
to hold at load.

`SearchEngine::from_json_validated` adds a checked JSON loader, and the bundled
Gemini specs now use it. But every Python placement site builds its engine from
an already-parsed `ArchSpec`, so four paths still took one unchecked:
`SearchEngine.from_arch_spec`, the entropy solver constructor, the target
generator DSL, and `PolicyRunner.from_arch_spec`. A user loading a custom
architecture via the unvalidated `ArchSpec.from_json` reached the search
regardless.

`from_arch_spec` now validates and returns `Result`; it had a single caller, so
making the checked form the default cost nothing. `from_index` stays as the
documented escape hatch. The binding paths that build a `LaneIndex` directly go
through one `validated_lane_index` helper, so a future entry point has an
obvious thing to call rather than another chance to forget.

Confirmed on the real path: a spec with a self-looping site bus now raises
`ArchSpecError` with the individual problems in its `errors` list, where it
previously produced a working engine.

## Verification

Both benchmark suites report **no differences**, which is expected rather than a
null result: the bundled specs are endpoint-disjoint, so no chain candidate can
arise and the new code path is never entered. The win shows up on specs with
overlapping buses, where it measured roughly a 20% reduction in move count on a
98-atom Ising kernel.

- 21 Rust test targets, 1831 Python tests
- clippy and fmt clean on the CI-gated crates; pyright clean
- `just benchmark-physical` and `just benchmark-logical`: no differences

The two tests that pinned the gap are flipped to assert assembly, joined by
coverage of the limits: three-hop chains, 2x2 and 3x2 cascades, entry-order
independence, blocked heads rejecting the whole chain, and misaligned chains.

Three further tests cover a corner worth naming, since repair works in
*position* space: `pos_to_src` deliberately tolerates two sources at one grid
position (last-insert-wins — the `full.json` fixture stacks every word at
identical coordinates). When another source won the position, a repair pulls in
coordinates that resolve to the winner, the blocking occupant never becomes a
rectangle source, and the same repair is requested from an unchanged rectangle.
The `grew` guard abandons the rectangle instead of spinning, and rollback undoes
every inserted coordinate — verified load-bearing by mutation (dropping the
repair-rollback leaves `xs={1}` and fails the test).

## Note on scope

An earlier revision also added a post-hoc pass that merged adjacent move layers.
It was dropped: the serialisation that motivated it turned out to come from the
benchmark architecture splitting one rigid displacement across 16 buses, not
from the compiler. Fixing the spec removed the need, and dropping the pass also
removed two review findings it had introduced.

### DIFF
--- a/crates/bloqade-lanes-bytecode-python/src/errors.rs
+++ b/crates/bloqade-lanes-bytecode-python/src/errors.rs
@@ -39,6 +39,25 @@ fn arch_spec_error_to_py(py: Python<'_>, error: &ArchSpecError) -> PyResult<PyOb
 }
 
 /// Convert a Vec<ArchSpecError> to a single Python ArchSpecError with an errors list.
+/// Build a [`LaneIndex`] from a Python-supplied [`ArchSpec`], rejecting a spec
+/// that fails structural validation.
+///
+/// Every path that turns a user-supplied spec into search machinery goes
+/// through here. The search layers treat per-bus acyclicity and endpoint
+/// uniqueness (#874) as *given* rather than checking them: rectangle growth
+/// exempts an occupied destination whose occupant moves in the same shot,
+/// which is what makes conveyor chains legal — and on a cyclic bus is exactly
+/// what would make a physically impossible rotation look legal too.
+pub fn validated_lane_index(
+    py: Python<'_>,
+    arch_spec: &bloqade_lanes_bytecode_core::arch::types::ArchSpec,
+) -> PyResult<bloqade_lanes_search::primitives::lane_index::LaneIndex> {
+    arch_spec
+        .validate()
+        .map_err(|errors| arch_spec_errors_to_py(py, errors))?;
+    Ok(bloqade_lanes_search::primitives::lane_index::LaneIndex::from_arch_spec(arch_spec))
+}
+
 pub fn arch_spec_errors_to_py(py: Python<'_>, errors: Vec<ArchSpecError>) -> PyErr {
     let module = match py.import(EXCEPTIONS_MODULE) {
         Ok(m) => m,
--- a/crates/bloqade-lanes-bytecode-python/src/search_python.rs
+++ b/crates/bloqade-lanes-bytecode-python/src/search_python.rs
@@ -1315,11 +1315,32 @@ impl PySearchEngine {
         }
     }
 
-    /// Create a ``SearchEngine`` from an ArchSpec JSON string.
+    /// Create a ``SearchEngine`` from an ArchSpec JSON string, **without
+    /// validating the spec**.
+    ///
+    /// Prefer :meth:`from_json_validated`. The search layers assume per-bus
+    /// acyclicity and endpoint uniqueness rather than checking them, so an
+    /// unvalidated spec with a cyclic bus would be routed as if a rotation
+    /// were a legal AOD operation.
     #[staticmethod]
     fn from_json(arch_spec_json: &str) -> PyResult<Self> {
         let engine = SearchEngine::from_json(arch_spec_json)
             .map_err(|e| PyValueError::new_err(format!("invalid arch spec JSON: {e}")))?;
+        Ok(Self {
+            inner: Arc::new(engine),
+        })
+    }
+
+    /// Create a ``SearchEngine`` from an ArchSpec JSON string, rejecting a
+    /// spec that fails structural validation.
+    ///
+    /// Raises ``ArchSpecError`` (with every individual problem in its
+    /// ``errors`` list) for an invalid spec, or ``ValueError`` for malformed
+    /// JSON.
+    #[staticmethod]
+    fn from_json_validated(arch_spec_json: &str, py: Python<'_>) -> PyResult<Self> {
+        let engine = SearchEngine::from_json_validated(arch_spec_json)
+            .map_err(|e| crate::errors::arch_spec_load_error_to_py(py, &e))?;
         Ok(Self {
             inner: Arc::new(engine),
         })
--- a/crates/bloqade-lanes-bytecode-python/src/search_python.rs
+++ b/crates/bloqade-lanes-bytecode-python/src/search_python.rs
@@ -577,8 +577,9 @@ impl PyEntropyScorer {
         beta: f64,
         gamma: f64,
         w_t: f64,
+        py: Python<'_>,
     ) -> PyResult<Self> {
-        let index = LaneIndex::from_arch_spec(&arch_spec.inner);
+        let index = crate::errors::validated_lane_index(py, &arch_spec.inner)?;
 
         let targets: Vec<(u32, u64)> = target
             .iter()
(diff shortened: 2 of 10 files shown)


weinbe58:
Thanks — addressed two of the three in cd33cd6d.

**Rollback allocation (`aod_grid.rs`) — fixed, and the comment was wrong too.**
Correct catch. `added_x.push(x)` fired whenever a new coordinate was inserted,
which is the common case, so every `try_add_point` call heap-allocated in the
hottest loop of candidate generation. The comment directly above it claimed the
buffers stayed unallocated unless a chain needed repairing, which was simply
false for the requested point.

Fixed as suggested: the requested point is tracked in a `bool` again, exactly
as the pre-repair code did, so the allocation-free rollback #882 tuned for is
preserved. The `Vec`s now record only *repair-induced* inserts, which cannot
occur on an endpoint-disjoint bus — so the shipped specs really do pay nothing,
and the comment now matches the code.

**Type stub (`_native.pyi`) — fixed.** `from_json_validated` is now declared on
`SearchEngine`, and `from_json` carries the same warning as the Rust docstring
about routing an unvalidated spec as if a rotation were a legal AOD operation.

**`repairs.as_deref_mut()` — not a defect.** `Option::as_deref_mut` requires the
*inner* type to implement `DerefMut`. Here the inner type is `&mut Vec<u64>`,
and `&mut T: DerefMut<Target = T>`, so the result is `Option<&mut Vec<u64>>`,
not `Option<&mut [u64]>` — one deref level, not two. `out` is a `&mut Vec<u64>`
and `push` is valid. The crate compiles and its 381 tests pass, which settles
it either way.

Verified after the fix: 381 search tests, clippy and fmt clean, pyright clean.

PR #901, opened 2026-08-08
fix(search): batch conveyor chains in the push-rotate condenser

### LINKED ISSUE #892: push_rotate::schedule forbids chain batching on a stale bus-disjointness assumption
## Summary

`push_rotate::schedule` — the condenser that turns the planner's sequential move list into batched AOD operations — builds **strict** precedence edges, and its stated justification no longer holds. From the module docs (`push_rotate/schedule.rs:16-20`):

> The edges are *strict* — a move cannot enter a vertex in the same operation that another leaves it. That would need the vertex to be both a source and a destination of one bus group, and **every bus has disjoint source and destination sets** (see the `feasibility` module in the open crate for why that property holds and why it matters).

That was true when written. It is not true now:

- #874 made the load-bearing invariant per-bus **acyclicity**, not endpoint disjointness, and `ArchSpec::validate` enforces exactly that.
- #866 established the uniform destination rule: an occupied destination is legal when its occupant vacates in the *same* group. A vertex being both a source and a destination of one operation is precisely a legal conveyor chain.
- #887 made the grid layer assemble those chains.
- The feasibility module the comment cites has itself been updated — `validate_bus_disjointness` is gone, replaced by the acyclicity check.

## Impact

On an endpoint-disjoint spec this is invisible: no vertex can be both a source and a destination of one bus, so the strict edges never over-constrain anything. On an **overlapping-bus spec** — which is now legal, and which at least one real architecture uses deliberately to improve move-option density — the scheduler forbids exactly the batching that makes conveyor chains worthwhile, and serialises them into one operation per hop.

This is independent of the generator-side serialisation fixed in #887: a plan can be routed through Push and Rotate and be re-serialised here.

## Scope

- [ ] Relax the precedence edges: a move may enter a vertex in the same operation that another leaves it, provided both moves are in the same bus group and the resulting batch validates. The vacating move must still be in the *same* operation — sequencing across operations keeps its strict edge.
- [ ] The batching step already re-validates every emitted batch with `ArchSpec::check_lanes`, which is the authoritative geometry rule; the occupancy half now wants `AtomStateData::validate_moves` (the canonical rule from #866), so consider validating batches against replayed state rather than geometry alone.
- [ ] Correct the module docs, which currently assert disjointness as a property of every bus.
- [ ] Test on an overlapping-bus fixture — `crate::test_utils::chain_arch_json()` exists for exactly this, and `tests/push_rotate.rs` currently asserts endpoint-disjointness of its own output (`:101-111`, `:186-190`), an assertion that will need revisiting since chains legitimately violate it.

## Related

- #866 — execution-model contract (uniform destination rule)
- #874 — per-bus acyclicity replaces disjointness as the invariant
- #887 — chain assembly in the grid layer; same class of stale assumption, fixed one layer up

weinbe58:
Closes #892.

`push_rotate::schedule` built **strict** precedence edges everywhere, justified by a comment asserting that "every bus has disjoint source and destination sets". That is no longer true — #874 made per-bus *acyclicity* the load-bearing invariant, #866 established the uniform destination rule, and the `validate_bus_disjointness` check the comment pointed at has itself been removed. On an overlapping-bus spec the strict edges forbade exactly the batching that makes conveyor chains worthwhile, serialising them into one operation per hop. Same class of stale assumption as #887, one layer down.

## What changed

**Relaxed the precedence edges.** A vertex edge is now classified strict or **chain** — chain when the predecessor *vacates* the vertex the successor *enters*, both lanes share a bus group, and they are different agents (one atom cannot ride two lanes at once). A chain dependency means "the same operation or a later one, never an earlier one": a batch may include such a move only if every outstanding chain dependency rides with it. A strict edge between the same pair always wins.

**Rectangles carry their chain dependencies.** A chain dependency can only be met by another cell of the same operation, so a rectangle that would strand a chain leader is trimmed a column at a time — the AOD drives a complete Cartesian product and cannot compromise on rectangularity. Rows are left alone, since every row set is enumerated separately anyway.

**Batches are gated by the execution model, not geometry alone.** `ArchSpec::check_lanes` is replaced by `AtomStateData::validate_moves` + `apply_validated` against a replayed placement — the canonical rule from #866, both halves. Mover-only occupancy is sufficient: the planner only steps onto a vertex it has established is empty, so a stationary atom can never sit on a move's destination. The argument is written up on `mover_state`.

**Docs corrected.** The disjointness claim and its dead reference in `schedule.rs`, plus `push_rotate/mod.rs`, which asserted a legal AOD move is "vertex-disjoint single moves into empty destinations".

## Tests

Six new unit tests on `test_utils::chain_arch_json()`, the overlapping-bus fixture. Four fail on the old strict edges — `[1, 1, 1]` operations where the fix gives `[3]` — and `a_rectangle_that_would_strand_a_chain_leader_is_trimmed` fails with the trim disabled (`[1, 3, 1]`: the illegal 2×2 gets picked, validation rejects it, the fallback degrades), which pins both the trim and the safety net. `the_router_batches_a_chain_into_one_operation` goes through `solve_push_rotate`, so it also clears `assert_move_layers_executable` replaying from the *full* root placement.

`tests/push_rotate.rs:101-111` and `:186-190` asserted the router's own output was endpoint-disjoint. Chains legitimately violate that, and it only ever held because both fixtures are disjoint specs — removed, with the module doc now stating why `validate_moves` is the actual rule.

## The shipped specs are provably unaffected

Both Gemini specs have zero src/dst overlap on every bus, so no vertex can be both a source and a destination of one bus group, no chain edge can form, `needs` stays empty and the trim early-returns. Verified empirically too: `examples/headroom` output is byte-identical to `main` — `scheduler_ops` 49 / 94 / 163 at k=4 / 8 / 16. Push and Rotate is not in the benchmark registry, so **no baselines move**.

## Verification

- `cargo test -p bloqade-lanes-search` — 390 lib tests, 22 push-rotate integration tests, green
- `cargo test --workspace` — green
- `cargo clippy -p bloqade-lanes-search --all-targets -- -D warnings` — clean
- `uv run pytest python/tests` — 1831 passed, 9 skipped, 1 xpassed

## Note for follow-up

While here I prototyped and measured **filler lanes** (admitting empty-source cells so a rectangle can span columns a mover is missing from) and did *not* ship it. On the shipped specs the binding constraint is bus-group mismatch, not rectangle geometry: physical k=16 has ~6.95 moves ready per step but the largest same-bus-group cohort averages **1.36**, because 44 bus groups outnumber ready moves ~6:1. Fillers measured −1% to −6.7% operations on physical and **0.0% on logical**, for +9% atom-free lanes and a signature change to `schedule`. The upstream lever is `PlanHeuristics::score_step` alignment — `headroom` puts DAG depth at 32 against 163 operations. Worth revisiting only once a wide-bus spec exists.

### DIFF
--- a/crates/bloqade-lanes-search/src/push_rotate/schedule.rs
+++ b/crates/bloqade-lanes-search/src/push_rotate/schedule.rs
@@ -13,11 +13,22 @@
 //! precedence edge. Together with each agent's own move order that is
 //! sufficient for validity: a move's destination is vacated before it enters.
 //!
-//! The edges are *strict* — a move cannot enter a vertex in the same operation
-//! that another leaves it. That would need the vertex to be both a source and
-//! a destination of one bus group, and every bus has disjoint source and
-//! destination sets (see the `feasibility` module in the open crate for why
-//! that property holds and why it matters).
+//! Most of those edges are **strict**: the successor's operation must come
+//! after the predecessor's. One pattern is not. When the predecessor *vacates*
+//! the vertex the successor *enters*, and both moves ride the same bus group,
+//! the execution model's uniform destination rule (issue #866) makes the pair
+//! legal in one operation — that is precisely a conveyor chain hop. Those are
+//! recorded as **chain** dependencies, meaning "the same operation or a later
+//! one, never an earlier one": a batch may include such a move only if every
+//! outstanding chain dependency of it is in the same batch.
+//!
+//! Chain dependencies only arise where one bus's source and destination sets
+//! overlap. That is legal — per-bus acyclicity, not endpoint disjointness, is
+//! the invariant `ArchSpec::validate` enforces (issue #874) — and at least one
+//! architecture uses it deliberately to widen its move options. The shipped
+//! Gemini specs do keep their endpoints disjoint, and there no vertex can be
+//! both a source and a destination of one bus group, so every edge is strict
+//! and this scheduler behaves exactly as it did before the relaxation.
 //!
 //! ## Batching
 //!
@@ -25,14 +36,18 @@
 //! `(move_type, bus_id, zone_id, direction)` group whose source positions form
 //! a complete X×Y rectangle — the Cartesian product of their distinct x and y
 //! values. So each scheduling step takes the ready set, partitions it by bus
-//! group, and picks the largest rectangle available. Every emitted batch is
-//! re-validated with `ArchSpec::check_lanes`, which is the authoritative rule.
+//! group, and picks the largest rectangle that carries its chain dependencies
+//! with it. Every emitted batch is then re-validated with
+//! `AtomStateData::validate_moves` against the replayed placement — the
+//! canonical executability check, `ArchSpec::check_lanes` geometry *plus* the
+//! occupancy rules — and a group it rejects degrades to a single move.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 
 use crate::feasibility::graph::{LaneGraph, VertexId};
 use crate::primitives::lane_index::LaneIndex;
 use bloqade_lanes_bytecode_core::arch::addr::{LaneAddr, LocationAddr};
+use bloqade_lanes_bytecode_core::atom_state::AtomStateData;
 
 use crate::push_rotate::context::GroupKey;
 use crate::push_rotate::state::Move;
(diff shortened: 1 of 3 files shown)


crates/bloqade-lanes-search/src/push_rotate/schedule.rs
```diff
@@ -175,49 +179,221 @@ pub fn schedule_with(
...
+                .copied()
+                .find(|i| !needs.contains_key(i))
+                .unwrap_or(solo);
+            batch = vec![one];
+            batch_lanes = vec![lanes[one]];
+            validated = state.validate_moves(&batch_lanes, arch).ok();
+        }
+        // A single move the execution model refuses means the plan itself does
+        // not execute — a planner bug, not something to schedule around.
+        state = state.apply_validated(&validated?).ok()?;
```
weinbe58: one issue I see here is that if that if there is some kind of a bug in the generation of a batch then the fallback will always catch the bug and will silently fail. Its probably better to make this a error out instead

weinbe58: Agreed, and fixed in de3849ff — the fallback is gone; an unvalidatable batch now panics naming the batch and the errors, matching what `search::verify::assert_move_layers_executable` does for a packaged plan and for the same stated reason.

Two things I checked before changing it:

**The fallback was dead code.** I ran the whole suite with that branch replaced by a panic — both Gemini specs, walk and permutation instances, the 120-instance carved brute-force sweep, the chain fixtures — and it is never reached. Its only possible effect was the one you describe.

**The assertion catches the class of bug it exists for.** Disabling `close_rectangle` (the chain trim) now aborts with

```
the condenser built an operation the execution model rejects (this is a bug in
the scheduler, not in the request): 4 of 5 moves, 4 lanes
  - lane LaneAddr { .. site_id: 1, bus_id: 0 } targets LocationAddr { .. site_id: 2 },
    which is occupied by qubit 0 that does not move in this group
```

where it previously emitted a quietly serialised `[1, 3, 1]` schedule that only an exact batch-size assertion in one test would have caught.

Worth noting this is strictly a net reduction in silence: since these batches flow into `assert_move_layers_executable` via `solve_push_rotate` anyway, an illegal batch already panicked downstream — just with a vaguer message, and only on that path. Direct `schedule()` callers got the silent degrade.

crates/bloqade-lanes-search/src/push_rotate/schedule.rs
```diff
@@ -13,26 +13,41 @@
...
 //! `(move_type, bus_id, zone_id, direction)` group whose source positions form
 //! a complete X×Y rectangle — the Cartesian product of their distinct x and y
 //! values. So each scheduling step takes the ready set, partitions it by bus
-//! group, and picks the largest rectangle available. Every emitted batch is
-//! re-validated with `ArchSpec::check_lanes`, which is the authoritative rule.
+//! group, and picks the largest rectangle that carries its chain dependencies
+//! with it. Every emitted batch is then re-validated with
+//! `AtomStateData::validate_moves` against the replayed placement — the
+//! canonical executability check, `ArchSpec::check_lanes` geometry *plus* the
+//! occupancy rules — and a group it rejects degrades to a single move.
```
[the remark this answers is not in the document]
weinbe58: Correct, thanks — the docs and the code disagreed. Resolved in de3849ff by changing the code rather than the docs, since @weinbe58 raised the underlying design point in parallel: the degrade-to-a-single-move fallback is gone entirely, so a rejected batch now panics instead of quietly producing a smaller operation.

That also removes the second meaning `None` had acquired. `schedule` again returns `None` only when a move has no lane realising it — the one case its doc comment claims — so the caller ambiguity you flagged (plan bug vs. resolvable batching) no longer exists. The module docs now state that the validation is an assertion, not a fallback, and why.

PR #931, opened 2026-08-17
fix: added hintconst to fix bug for deserialization

jasonhan3:
# Restore logical-result post-processing after kernel serialization

Closes QuEraComputing/bloqade-internal#406.

## Summary

- Rebuild Kirin constant hints before generating logical-result post-processing.
- Add a regression test covering the stored-program JSON path used by
  `GeminiLogicalResult.logical_results()`.

## Problem

Kirin JSON serialization intentionally does not preserve SSA `const` hints.
After decoding a stored logical kernel, measurement-ID analysis therefore could
not resolve literal indexing such as `measurements[0][0]`. Post-processing was
silently unavailable, and `logical_results()` returned raw measurement arrays
instead of the kernel's logical result values.

## Fix

`generate_post_processing()` now runs `HintConst` before `MeasurementIDAnalysis`.
This recreates the derived constant hints from decoded IR, allowing the analysis
to resolve terminal-measurement indexing correctly.

## Testing

- Added an end-to-end regression that serializes a detector-returning logical
  kernel, stores its program JSON, and verifies that `logical_results()` yields
  detector values rather than raw shots.
- `uv run pytest python/tests/gemini/test_post_processing.py python/tests/gemini/device/test_result.py`
- `uv run black --check ...`
- `uv run isort --check-only ...`
- `uv run pyright ...`


python/bloqade/gemini/post_processing.py
```diff
@@ -65,6 +66,11 @@ def generate_post_processing(
 
     """
 
+    # JSON serialization deliberately omits SSA hints. Rebuild constant hints
+    # before analysing measurement indexing so a decoded kernel can resolve
+    # expressions such as ``measurements[0]``.
+    HintConst(mt.dialects, no_raise=False)(mt)
```
weinbe58: I would use a `Fold` pass instead. this is more idiomatic way to get the hints.

weinbe58 (changes requested):

python/bloqade/gemini/post_processing.py
jasonhan3: I think using a `Fold` pass causes errors because the deserialized kernel doesn't store `uses` correctly

```
original py.ilist.new result: uses = 1
decoded  py.ilist.new result: uses = 0
```

so the dead code elimination pass will incorrectly remove some SSA values which will cause problems

weinbe58: this should be fixed in newer versions of Kirin

jasonhan3: Just pushed a change to use `Fold`; I also bumped the dependent version of `kirin-toolchain` in `pyproject.toml`, so I'm not sure if this change can be backported?

weinbe58 (approved):

PR #934, opened 2026-08-19
fix(demo): correct GHZ preparation in ghz_moves_demo

### LINKED ISSUE #930: [Demo fix] Wrong circuit in the `ghz_moves_demo.py`
## Summary

The [`ghz_optimal`](https://github.com/QuEraComputing/bloqade-lanes/blob/e77c46fc51a46236423a34026144b32884d127c0/demo/ghz_moves_demo.py#L29-L39) demo kernel does not prepare the 10-qubit GHZ state

$$
(|0000000000\rangle + |1111111111\rangle) / \sqrt{2}.
$$

Exact Cirq statevector simulation gives uniform support on 128 computational-basis states. The `log_depth_ghz` kernel in the same demo produces the expected state. This is a correctness issue in the demo circuit rather than the movement visualizer. The movement output can look reasonable while the kernel prepares the wrong quantum state.

**To prevent this type of isse:** Use a simulator to sample and plot the measurement counts. Calculate the Hellinger fidelity with the ideal counts and make sure it is close to 1.

## Reproduce the issue

```python
import cirq
import numpy as np

from bloqade import cirq_utils, qubit, squin


@squin.kernel(typeinfer=True, fold=True)
def ghz_optimal():
    qs = qubit.qalloc(10)
    squin.broadcast.sqrt_y(qs)
    squin.z(qs[0])
    squin.cz(qs[0], qs[5])
    squin.broadcast.cz(
        ilist.IList([qs[0], qs[5]]), ilist.IList([qs[1], qs[6]])
    )
    squin.broadcast.cz(qs[:2] + qs[5:7], qs[2:4] + qs[7:9])
    squin.broadcast.cz(
        ilist.IList([qs[3], qs[8]]), ilist.IList([qs[4], qs[9]])
    )
    squin.broadcast.sqrt_y(qs)
    squin.sqrt_y_adj(qs[0])

# Convert to cirq circuits
circuit = cirq_utils.emit_circuit(ghz_optimal)
state = cirq.Simulator(dtype=np.complex128).simulate(circuit).final_state_vector
probabilities = np.abs(state) ** 2
support = np.flatnonzero(probabilities > 1e-12)

print(f"nonzero basis states: {len(support)}")
print(f"P(0000000000): {probabilities[0]:.6f}")
print(f"P(1111111111): {probabilities[-1]:.6f}")
```

Observed:

```text
nonzero basis states: 128
P(0000000000): 0.000000
P(1111111111): 0.000000

```

Expected:

```text
nonzero basis states: 2
P(0000000000): 0.500000
P(1111111111): 0.500000
```

## Cause

Every qubit receives an initial `sqrt_y`. After a target qubit receives its incoming CZ, it needs `sqrt_y_adj` before that qubit is reused as a control in the next CZ layer. The current implementation defers the remaining single-qubit rotations to one global pulse at the end. Those rotations do not commute through the later CZ gates, so the resulting tree-graph-state circuit is not equivalent to the intended GHZ preparation.

## Proposed fix

Apply `sqrt_y_adj` to each layer's newly connected targets immediately after that layer, and remove the final global rotations:

```diff
     squin.z(qs[0])
     squin.cz(qs[0], qs[5])
+    squin.sqrt_y_adj(qs[5])
     squin.broadcast.cz(ilist.IList([qs[0], qs[5]]), ilist.IList([qs[1], qs[6]]))
+    squin.broadcast.sqrt_y_adj(ilist.IList([qs[1], qs[6]]))
     squin.broadcast.cz(qs[:2] + qs[5:7], qs[2:4] + qs[7:9])
+    squin.broadcast.sqrt_y_adj(qs[2:4] + qs[7:9])
     squin.broadcast.cz(ilist.IList([qs[3], qs[8]]), ilist.IList([qs[4], qs[9]]))
-    squin.broadcast.sqrt_y(qs)
-    squin.sqrt_y_adj(qs[0])
+    squin.broadcast.sqrt_y_adj(ilist.IList([qs[4], qs[9]]))
```

With this change, the same exact statevector check reports:

```text
nonzero basis states: 2
P(0000000000): 0.500000
P(1111111111): 0.500000
```

The fixed kernel also produces only `0000000000` and `1111111111` when compiled and sampled noiselessly through `PhysicalSimulator`.

## Environment

- Confirmed that the affected function is still present on upstream `main` at commit [`e77c46fc51a46236423a34026144b32884d127c0`](https://github.com/QuEraComputing/bloqade-lanes/commit/e77c46fc51a46236423a34026144b32884d127c0) (2026-08-17).
- Python 3.12.13
- `bloqade-lanes==0.11.0.dev0`
- `bloqade-circuit==0.14.4`
- `kirin-toolchain==0.22.11`
- `cirq-core==1.7.0`
- `numpy==2.2.6`
- macOS 26.6, Apple Silicon

## Related context

PR [#173](https://github.com/QuEraComputing/bloqade-lanes/pull/173) previously completed the last CZ layer of the binary tree. This report concerns the remaining placement of the single-qubit basis changes.

weinbe58:
## Summary

Fixes the `ghz_optimal` kernel in [`demo/ghz_moves_demo.py`](demo/ghz_moves_demo.py), which prepared the **wrong quantum state** (reported in #930).

The original kernel deferred every basis-change rotation to a single global `sqrt_y` pulse at the end. Those rotations do not commute through the intervening CZ layers, so the kernel produced a tree graph state — uniform support over 128 computational-basis states — instead of the intended 10-qubit GHZ state $(|0\cdots0\rangle + |1\cdots1\rangle)/\sqrt{2}$.

The fix (as proposed in the issue) applies `sqrt_y_adj` to each layer's newly connected targets immediately after that layer, and removes the trailing global rotations.

## Verification

Using `bloqade.lanes.utils.check_circuit` to compare state vectors against a reference GHZ (textbook `H` + `CX` chain — a GHZ state is permutation-symmetric, so qubit ordering doesn't matter):

| kernel | nonzero basis states | P(0…0) | P(1…1) | `check_circuit` vs reference GHZ |
|---|---|---|---|---|
| reference GHZ (H + CX chain) | 2 | 0.500000 | 0.500000 | — |
| `ghz_optimal` (original) | 128 | 0.000000 | 0.000000 | ❌ `False` |
| `ghz_optimal` (fixed) | 2 | 0.500000 | 0.500000 | ✅ `True` |

The fixed kernel is state-vector-equivalent (up to global phase) to a correct GHZ preparation.

Closes #930

### DIFF
--- a/demo/ghz_moves_demo.py
+++ b/demo/ghz_moves_demo.py
@@ -32,11 +32,13 @@ def ghz_optimal():
     squin.broadcast.sqrt_y(qs)
     squin.z(qs[0])
     squin.cz(qs[0], qs[5])
+    squin.sqrt_y_adj(qs[5])
     squin.broadcast.cz(ilist.IList([qs[0], qs[5]]), ilist.IList([qs[1], qs[6]]))
+    squin.broadcast.sqrt_y_adj(ilist.IList([qs[1], qs[6]]))
     squin.broadcast.cz(qs[:2] + qs[5:7], qs[2:4] + qs[7:9])
+    squin.broadcast.sqrt_y_adj(qs[2:4] + qs[7:9])
     squin.broadcast.cz(ilist.IList([qs[3], qs[8]]), ilist.IList([qs[4], qs[9]]))
-    squin.broadcast.sqrt_y(qs)
-    squin.sqrt_y_adj(qs[0])
+    squin.broadcast.sqrt_y_adj(ilist.IList([qs[4], qs[9]]))
 
 
 mt = LogicalPipeline().emit(log_depth_ghz)


jasonhan3 (approved):
LGTM

PR #942, opened 2026-08-21
feat: first pass at impl result postproc

jasonhan3:
- Added `measurements`, `detectors`, `observables`, `return_values` to `GeminiLogicalResult`
- Added postprocessing from 160 SLM sites to measurements

### DIFF
--- a/python/bloqade/gemini/device/logical/result.py
+++ b/python/bloqade/gemini/device/logical/result.py
@@ -62,7 +62,8 @@ class GeminiLogicalResult(Result, Generic[RetType]):
             kernel_json = program["content"]
             kernel_mt = logical.kernel.decode_json(kernel_json)  # type: ignore[attr-defined]
             slm_to_raw, postprocessing_function = get_slm_mapping_postprocessing(
-                kernel_mt
+                kernel_mt,
+                invert_bits=True,
             )
             # NOTE: what if err in postproc fn generation/errors here??
             postprocessing_functions[idx] = (slm_to_raw, postprocessing_function)
--- a/python/bloqade/gemini/device/logical/result.py
+++ b/python/bloqade/gemini/device/logical/result.py
@@ -1,18 +1,26 @@
-from collections.abc import Callable
+from collections.abc import Callable, Sequence
 from dataclasses import dataclass
-from typing import TypeVar
+from functools import cached_property
+from typing import Generic, TypeVar, overload
 
 import numpy as np
 from bloqade.core.device import Result
 
 from bloqade.gemini import logical
 from bloqade.gemini.post_processing import generate_post_processing
+from bloqade.lanes.analysis.atom import PostProcessing
+
+from .utils import (
+    aligned_detected_and_sorted_shots_for_subtasks,
+    get_slm_mapping_postprocessing,
+    shot_results_for_subtasks,
+)
 
 RetType = TypeVar("RetType")
 
 
 @dataclass(kw_only=True)
-class GeminiLogicalResult(Result):
+class GeminiLogicalResult(Result, Generic[RetType]):
     """Result view over stored Gemini logical shots.
 
     Merge-oriented methods assume each selected task ID has the same subtask
@@ -26,7 +34,13 @@ class GeminiLogicalResult(Result):
             subtask scope. Defaults to the DETECTED frame type.
     """
 
-    def postprocessing_functions(self) -> dict[int, Callable | None]:
+    @cached_property
+    def _slm_postprocessing_functions(
+        self,
+    ) -> dict[
+        int,
+        tuple[Callable[[np.ndarray], list[list[bool]]], PostProcessing[RetType]],
+    ]:
         """Decode stored programs and build post-processing functions.
 
         Program records are scoped by `shot_filter.task_ids`. When multiple
@@ -46,11 +60,64 @@ class GeminiLogicalResult(Result):
                 # NOTE: merging across task_ids means we assume all of them identical
                 continue
             kernel_json = program["content"]
-            kernel_mt = logical.kernel.decode_json(kernel_json)
-            postprocessing_function = generate_post_processing(kernel_mt)
-            postprocessing_functions[idx] = postprocessing_function
+            kernel_mt = logical.kernel.decode_json(kernel_json)  # type: ignore[attr-defined]
+            slm_to_raw, postprocessing_function = get_slm_mapping_postprocessing(
+                kernel_mt
+            )
+            # NOTE: what if err in postproc fn generation/errors here??
+            postprocessing_functions[idx] = (slm_to_raw, postprocessing_function)
 
         return postprocessing_functions
+
+    def postprocessing_functions(self) -> dict[int, Callable | None]:
+        """Build legacy compact-shot postprocessors from stored programs.
+
+        This method preserves the original ``GeminiLogicalResult`` API for
+        result stores whose bitstrings are already in compact physical
+        measurement order. Raw 160-site SLM frames use
+        ``_slm_postprocessing_functions`` instead.
+        """
+        task_ids = self.shot_filter.task_ids
+        programs = self.storage.get_programs(task_ids=task_ids)
+        postprocessing_functions = {}
+        for program in programs:
+            idx = program["program_index"]
+            if idx in postprocessing_functions:
+                continue
+            kernel_mt = logical.kernel.decode_json(program["content"])  # type: ignore[attr-defined]
+            postprocessing_functions[idx] = generate_post_processing(kernel_mt)
+
+        return postprocessing_functions
+
+    @cached_property
+    def measurements(self) -> Sequence[Sequence[Sequence[bool]]]:
+        ret_vals: list[list[list[bool]]] = []
+        # TODO: OK to set verify=True?
+        subtasks = self.subtasks(verify=True)
+        postprocessing_functions = self._slm_postprocessing_functions
+        shot_results = shot_results_for_subtasks(
+            self.storage, self.shot_filter, subtasks, frame_type="DETECTED"
+        )
+
+        for shot_result, subtask in zip(shot_results, subtasks):
+            func = postprocessing_functions[subtask["program_index"]][0]
+            ret_vals.append(func(shot_result))
+
+        return ret_vals
+
+    @overload
+    def logical_results(
+        self,
+        verify: bool = True,
+        postprocessing_functions: None = None,
+    ) -> list[list[RetType]]: ...
+
+    @overload
+    def logical_results(
+        self,
+        verify: bool,
+        postprocessing_functions: dict[int, Callable[[np.ndarray], RetType] | None],
+    ) -> list[RetType | np.ndarray]: ...
 
     def logical_results(
         self,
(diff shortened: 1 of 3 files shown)


python/bloqade/gemini/device/logical/utils.py
```diff
@@ -0,0 +1,214 @@
...
+    interpreter = atom.AtomInterpreter(
+        physical_move_kernel.dialects,
+        arch_spec=arch_spec,
+    )
+    frame, _ = interpreter.run(physical_move_kernel)
+
+    # Each MeasureResult identifies both:
+    # - its compact measurement-record column expected by post-processing
+    # - the physical SLM location at which it was measured.
+    records_by_id = {}
```
weinbe58: what's wrong with using the get_post_processing method here? Is the zone ordering wrong?

jasonhan3: I think `get_shot_remapping` assumes that the kernel returns an `IListResult[IListResult[MeasureResult]]`; but to accomodate kernels that return `default_post_processing(reg)`, I opted for a custom pass here.

But also there is a separate problem of reordering the SLM sites from the arch spec to that of the ordering from left to right, bottom to top (assumed to be returned back). I think the ordering of the SLM sites from the physical architecture is attached below (which is why I do a conversion of the SLM indices in this function).

Image

python/bloqade/gemini/device/logical/utils.py
```diff
@@ -0,0 +1,170 @@
...
+            return []
+        if zone0_shots.ndim != 2:
+            raise ValueError(f"Expected a 2-D array, got {zone0_shots.shape}.")
+        if zone0_shots.shape[1] != expected_zone0_sites:
+            raise ValueError(
+                f"Expected {expected_zone0_sites} Zone-0 columns, "
+                f"got {zone0_shots.shape[1]}."
+            )
+
+        measurement_shots = zone0_shots[:, mapping]
```
weinbe58, line 161: **Root cause is upstream in `get_shot_remapping`, not in this file** — filed as #967. Leaving the detail here since this is where it surfaces.

`mapping` comes from `AtomInterpreter.get_shot_remapping`, which walks the kernel's *return value* in row-major order and appends one Zone-0 index per `MeasureResult`. So `measurement_shots[:, k]` is the k-th element of the flattened return value.

But every `PostProcessing.emit_*` callable is built by `constructor_function`, which does `measurements[elem.measurement_id]` — and `measurement_id` is the *global* measurement-record index assigned per `GetFutureResult` in IR order (`impl.py:246`). The two agree only when the return value covers all measurement records starting at 0 in emission order.

Reproduced on this branch:

```python
@lg.kernel(aggressive_unroll=True)
def k3():
    q = lg.qalloc_at(ilist.IList([4, 5]))
    squin.h(q[0])
    m = lg.terminal_measure(q)
    return ilist.IList([m[1]])
```

14 measurement records exist, but `mapping == [80, 81, 82, 83, 84, 85, 86]` (7 entries), while `emit_return` reads `measurements[7..13]`:

```
emit_return raised: IndexError list index out of range
```

Silent-corruption variant — same count, permuted order:

```python
    return ilist.IList([m[1], m[0]])
```

`mapping == [80..86, 64..70]`. With only zone bits 80–86 set to `True`, the projected array is correctly `[[True]*7 + [False]*7]`, but `emit_return` re-applies the permutation and yields:

```
IList([IList([False]*7), IList([True]*7)])
```

i.e. the two logical blocks swapped, no error raised. (Returning the *first* subset, `[m[0]]`, happens to work by luck, which makes this easy to miss.)

Affects all five new properties (`measurements`, `return_values`, `detectors`, `observables`, `filling_at_start`).

---

**Correction to my original comment:** I first read this as a bug in this line. It isn't — the code here does exactly what the API documents:

- `_shot_remapping.py:16` — "produce a flat list of Zone-0 bitstring indices whose order **matches the per-measurement array consumed by `generate_post_processing`**"
- `_shot_remapping.py:64` — "**Index directly into the post-processing flat array.**"

The implementation appends in traversal order (`_shot_remapping.py:183`) and never reads `measurement_id`, so it doesn't honour that contract. Its own `ShotRemappingOk` docstring contains the contradiction in adjacent sentences. Nothing caught it because the test suite disclaims the field the contract depends on — `test_shot_remapping.py:71`: *"`get_shot_remapping` projects on `location_address` and ignores `measurement_id`, so a placeholder `0` record index is fine here."*

So switching to this API in e52924e was the right call; it just inherited a broken contract.

**What this means for this PR:** I don't think this needs to block you. #967 has the suggested fix (place indices by `measurement_id`, error when the ids aren't a contiguous `0..n-1` cover). Options here, your call:

1. Land as-is and let #967 fix the ordering underneath you — no change needed in this file.
2. Add a cheap guard now so the silent-corruption case can't ship: `assert len(mapping) == interpreter.measurement_record_count` after line 141. That converts the permuted-return case into a loud failure and costs nothing once #967 lands.

Sorry for the misattribution in the first pass.

python/bloqade/gemini/device/logical/device.py
```diff
@@ -19,6 +32,20 @@ class GeminiLogicalDevice(Device[GeminiLogicalFuture]):
     `Device` factory and defaults `context_name` to "gemini-logical".
     """
 
+    program_language: str = "squin"
+    language_version: str = _bloqade_version
+    validation_suite: ValidationSuite | None = field(
```
weinbe58, line 37: **`validation_suite` and `dialect_group` are never read.**

Nothing in the installed bloqade stack consults these. `Device.task()` / `batch_task()` / `parameter_scan()` build the task from `context_name`, `kernel`, `num_shots`, `arguments`, `metadata`, `program_language`, `language_version`, `future_cls`, `kernel_serializer` — no reference to `self.validation_suite` or `self.dialect_group`. `grep -rn validation_suite` over `site-packages/bloqade/` returns nothing outside this file.

Verified against bloqade-circuit 0.15.0:

```python
@lg.kernel(aggressive_unroll=True, verify=False)
def bad():
    q = lg.qalloc_at(ilist.IList([4, 5]))
    squin.t(q[0])          # non-Clifford: GeminiLogicalValidation should reject
    return lg.terminal_measure(q)

GeminiLogicalDevice().task(bad, num_shots=1)   # succeeds, no validation
```

The PR description says the device is now configured with "Gemini's language, dialect, and validation defaults", so a reader will reasonably assume invalid kernels are now caught at submission. Either wire the suite into a submission-time hook, or drop the fields until core supports them.

python/bloqade/gemini/device/logical/utils.py
```diff
@@ -0,0 +1,170 @@
...
+        arch_spec=arch_spec,
+    )
+    shot_mapping = interpreter.get_shot_remapping(physical_move_kernel)
+    if isinstance(shot_mapping, ShotRemappingErr):
+        raise ShotRemappingException(
+            f"Failed to produce shot remapping, error message: {shot_mapping.diagnostic.message}, reason: {shot_mapping.diagnostic.offending_value}"
+        )
+    mapping = shot_mapping.mapping
+    zone0_locations = list(arch_spec.yield_zone_locations(zone0))
+    print(f"row-major SLM mapping: {mapping}")
```
weinbe58: **Leftover debug `print` in library code.**

This fires on every mapping construction, with no way to suppress it. Reading `result.measurements` prints e.g.

```
row-major SLM mapping: [64, 65, 66, 67, 68, 69, 70, 80, 81, 82, 83, 84, 85, 86]
```

A 10-word program prints a 70-element list, and because `filling_at_start` builds a second cache entry with `invert_bits=False` (see the separate comment on `result.py`), the same kernel prints twice per result. In a notebook, or a batch with many programs, this floods the output.

Suggest dropping it, or moving it behind a logger at debug level.

python/bloqade/gemini/device/logical/result.py
```diff
@@ -1,94 +1,237 @@
...
+            list(
+                self._slm_postprocessing_functions()[subtask["program_index"]][
+                    1
+                ].emit_return(measurements)
+            )
+            for measurements, subtask in zip(self.measurements, self.subtasks())
+        ]
+
+    @cached_property
+    def detectors(self) -> Sequence[Sequence[Sequence[bool]]]:
```
weinbe58, line 182: **`detectors` / `observables` can't read back the canonical QEC program shape.**

All five new properties go through `get_shot_remapping`, which requires the return value to refine to `IListResult[IListResult[MeasureResult]]`. But `logical.default_post_processing` — the stdlib entry point that actually declares detectors and observables — returns a `(detectors, observables)` tuple, so it hits the `TupleResult` rejection path.

Net effect for a kernel ending in `return logical.default_post_processing(qubits)`: `measurements`, `return_values`, `detectors`, `observables`, and `filling_at_start` all raise `ShotRemappingException: outer return value did not refine to IListResult`. `test_default_postproc_remapping` encodes this as `xfail(strict=True)`.

Detectors do work for the narrower shape (`return terminal_measure(q)` with inline `squin.set_detector` calls) — I confirmed `emit_detectors` returns `[[True]]` there — but that isn't the shape the stdlib steers users toward. Worth calling out in the property docstrings as a known limitation, and worth a follow-up issue for tuple-returning kernels, since `detectors`/`observables` are the headline additions here.

python/bloqade/gemini/device/logical/device.py
```diff
@@ -19,6 +32,20 @@ class GeminiLogicalDevice(Device[GeminiLogicalFuture]):
     `Device` factory and defaults `context_name` to "gemini-logical".
     """
 
+    program_language: str = "squin"
+    language_version: str = _bloqade_version
```
weinbe58, line 36: **`language_version` has no effect on the submitted task.**

`Device.task()` / `batch_task()` / `parameter_scan()` take `language_version: str = "0.1.0"` as a *method parameter* and pass that to the task constructor; they never read `self.language_version`. Verified:

```
device.program_language = squin | device.language_version = 0.15.0
task.program_language      = squin
task.language_version      = 0.1.0        # <-- device field ignored
task.program_language_version = 0.15.0   # <-- from GeminiTaskMixin
```

So `GeminiLogicalDevice(language_version="1.2.3")` changes nothing, and the correct version only reaches storage because `GeminiTaskMixin.program_language_version` (in `task.py`) already overrides it. That also makes the `_bloqade_version = version("bloqade-circuit")` lookup at line 24 a duplicate of the one in `task.py`.

Same applies to `program_language` on line 35 — already set by `GeminiTaskMixin`. Suggest removing both fields (and the duplicate `version()` call) so there's one source of truth.

python/bloqade/gemini/device/logical/result.py
```diff
@@ -1,94 +1,237 @@
...
+        ]
+
+    @cached_property
+    def filling_at_start(self) -> Sequence[Sequence[Sequence[bool]]]:
+        """True indicates that the atom was present during the sorted frame; False indicates that it was not."""
+        from .utils import aligned_detected_and_sorted_shots_for_subtasks
+
+        ret_vals: list[list[list[bool]]] = []
+        subtasks = self.subtasks(verify=True)
+        postprocessing_functions = self._slm_postprocessing_functions(invert_bits=False)
```
weinbe58: **This recompiles every stored kernel a second time.**

`invert_bits=False` misses the `invert_bits=True` cache entry, so `_slm_postprocessing_functions` re-runs `get_slm_mapping_postprocessing` for every program — a full `LogicalPipeline(transversal_rewrite=True).emit()` plus two `AtomInterpreter` runs — purely to obtain a projection that skips the trailing `~`.

Reading both `measurements` and `filling_at_start` on a result with N programs therefore performs 2N logical→physical compilations. The mapping and the `PostProcessing` object are identical between the two conventions; only the final bit flip differs.

Suggest making inversion a parameter of the returned closure (`postprocess(shots, *, invert=True)`) or applying the `~` at the call site, and keying the cache on the kernel alone.

python/bloqade/gemini/device/logical/result.py
```diff
@@ -1,94 +1,237 @@
...
-            postprocessing_function = generate_post_processing(kernel_mt)
-            postprocessing_functions[idx] = postprocessing_function
+        for idx, kernel_json in self._program_contents_by_index().items():
+            kernel_mt = logical.kernel.decode_json(kernel_json)  # type: ignore[attr-defined]
+            postprocessing_functions[idx] = generate_post_processing(kernel_mt)
 
         return postprocessing_functions
 
+    @cached_property
+    def measurements(self) -> Sequence[Sequence[Sequence[bool]]]:
```
weinbe58, line 114: **`cached_property` over live storage: these can go stale.**

`measurements`, `return_values`, `detectors`, `observables`, and `filling_at_start` are all `cached_property` on a result *view* backed by mutable storage, whereas `logical_results()` — the method they're positioned to replace — re-reads storage on every call.

```python
res = future.partial_result()
print(res.measurements)      # say 100 of 1000 shots have landed
...
future.fetch()               # remaining 900 shots arrive
print(res.measurements)      # still the original 100 rows, silently
print(res.logical_results()) # would return all 1000
```

Not wrong if intended, but it's a behavioural difference from the API being deprecated and it's invisible at the call site. Either document that these snapshot at first access, or invalidate the cache when storage changes.

python/bloqade/gemini/device/logical/utils.py
```diff
@@ -0,0 +1,170 @@
...
+
+    arch_spec = physical.get_arch_spec()
+    physical_move_kernel = LogicalPipeline(transversal_rewrite=True).emit(sim_kernel)
+    zone0 = ZoneAddress(0)
+
+    interpreter = atom.AtomInterpreter(
+        physical_move_kernel.dialects,
+        arch_spec=arch_spec,
+    )
+    shot_mapping = interpreter.get_shot_remapping(physical_move_kernel)
```
weinbe58, line 136: **`no_raise` default hides the real failure.**

`get_shot_remapping` defaults to `no_raise=True`, so a genuine exception inside the atom analysis is caught by `Forward.run_no_raise` and reappears as the generic "outer return value did not refine to IListResult; got Bottom" — pointing the reader at the kernel's return type rather than the actual cause, with the original traceback gone. And since `_slm_postprocessing_functions` doesn't expose a knob, there's no way to re-run with `no_raise=False` from the public `GeminiLogicalResult` API.

Consider `no_raise=False` here (analysis crashes are compiler-pipeline regressions, which is exactly what you want to see), or threading an escape hatch through for debugging.

weinbe58 (commented):
Code review of the SLM post-processing path. I rebuilt the native extension on this branch and reproduced each of the findings below locally; the new test file passes (72 passed, 1 xfailed).

One blocker (**#1**): the projection into `measurement_shots` is ordered by the kernel's return-value traversal, while the `PostProcessing.emit_*` callables index by global `measurement_id`. Those two index spaces coincide only for kernels whose return value happens to cover every measurement record in emission order, so other kernels either crash or silently return permuted results.

The rest are correctness/clarity issues in the new device fields and helpers. Inline comments below, roughly most-severe first.

<sub>🤖 Automated review with [Claude Code](https://claude.com/claude-code), reviewed and posted by @pweinberg-quera</sub>

python/bloqade/gemini/device/logical/device.py
weinbe58: @jasonhan3 these are claude generated so I would appreciate just double checking this one in particular.

python/bloqade/gemini/device/logical/result.py
weinbe58: @jasonhan3 I think its worth double checking this because I'm not sure if the measurements are static, for example if you get partial results would you update this field in-place or just return a new result object?

python/bloqade/gemini/device/logical/utils.py
jasonhan3: I agree that this is a bug, but I think that this should be fixed in the later change when we modify the implementation of `get_shot_remapping`. I think that to address this bug, we have to change the implementation of `get_shot_remapping` to either (a) find the measurements from logical.terminal_measure or (b) enforce that the kernel returns qubits in allocation order (of the same length), both of which involve changing the implementation of `get_shot_remapping` (so I'm not sure if it's worth doing in this PR, if we are going to change `get_shot_remapping` in a future PR which solves this issue)

python/bloqade/gemini/device/logical/device.py
jasonhan3: Ah, these are read here: https://github.com/QuEraComputing/bloqade-core/blob/main/src/bloqade/core/device/device.py#L77-L86 in the abstract `Device` class that `GeminiLogicalDevice` inherits from; which are used in the implementation of `device.run_async(task: TaskBuilder)`: https://github.com/QuEraComputing/bloqade-core/blob/main/src/bloqade/core/device/device.py#L144

python/bloqade/gemini/device/logical/utils.py
jasonhan3: Thanks; just did

python/bloqade/gemini/device/logical/result.py
jasonhan3: Yes, this is right, but hopefully should be solved with the future PR that modifies `get_shot_remapping` to find the return values from `terminal_measure` as opposed from the kernel.

python/bloqade/gemini/device/logical/device.py
jasonhan3: `language_version` is needed because we need to pass it to `TaskBuilder._finalize() -> TaskDefinition` to create a `TaskDefinition`. https://github.com/QuEraComputing/bloqade-core/blob/main/src/bloqade/core/device/device.py#L144-L145

I agree that the value can be changed, but I think it's reasonable for now.  The duplication with `GeminiTaskMixin` is kind of expected because the tasks in "tasks.py" aren't expected to be used; we are opting for the TaskBuilder in bloqade-core and the device.run_async(task_builder) instead.

python/bloqade/gemini/device/logical/result.py
jasonhan3: Thanks; just addressed this to not do double compilation for the bit inversion

python/bloqade/gemini/device/logical/result.py
jasonhan3: Thanks; removed caching the measurements

python/bloqade/gemini/device/logical/utils.py
jasonhan3: ~~Set `no_raise` to be False to raise an exception~~

edit: leaving `no_raise` as default (True) as I already catch the shot remapping exception and raise it

weinbe58 (approved):
thanks!

Issue #940, opened 2026-08-21
pack_aod_rectangles lacks the chain closure, so Starlark policies still serialize packed chains


weinbe58:
Part of #887 — the known limit documented (not fixed) in #919.

`dsl::pipeline::pack_aod_rectangles` is the fourth caller of `BusGridContext::build_aod_grids`, and the only one that does not call `close_chain_entries` first. The heuristic generator and both entropy candidate paths close their mover sets over conveyor chains before building grids; a Starlark policy's candidate pipeline does not.

Consequence: cause 3 of #910 persists for the DSL pipeline on a chain-capable spec. `build_aod_grids` derives movers from `entries.keys()`, and #896's repair closure can only pull in cells whose source is already an entry (it must never promote a stationary atom, or the plan and the hardware disagree). A follower already sitting on its target is never nominated by scoring, so a policy's chain assembles only when every follower happens to be selected on its own merit — on a packed block it isn't, and the policy serializes or deadlocks where the built-in generators now solve.

## Fix shape

Mirror the entropy driver (`drivers/entropy.rs`): call `close_chain_entries` before `build_aod_grids`, and fold the returned `ChainLink`s into `entry_by_lane` so the co-selected follower moves are recorded in the emitted `PackedCandidate`s — an emitted plan must record every atom it moves.

## Tests

- The packed witness from #919 at the DSL-pipeline level: a policy on `crate::test_utils::chain_arch_json()` with the packed-block instance, asserting a one-shot chain candidate is emitted. The generator-level witnesses added in #919 are the template.
- No-op on endpoint-disjoint buses (the shipped specs) — structurally guaranteed since `close_chain_entries` cannot find a vacating occupant there, but worth pinning.
- Verify the new test fails with the closure call removed, per the discipline in #919.

Issue #938, opened 2026-08-21
Decide whether GreedyGenerator follows conveyor chains, or document it as chain-incapable


weinbe58:
Part of #887 (scope bullet 2 — the decision the other PRs left standing).

`GreedyGenerator` cannot express conveyor chains: `find_path_occupied` treats every atom as a wall, so the atom ahead on a conveyor bus is an obstacle rather than a follower. No candidate ever proposes a move into an occupied site, which means neither the grid-layer repair closure (#896) nor the selection-time chain closure (#919, `close_chain_entries`) can fire — the chain machinery is unreachable from this generator by construction.

#887 already concedes that documenting it as chain-incapable is a legitimate outcome; teaching it to follow chains would mean multi-agent pathfinding rather than single-atom BFS, which is effectively a different generator.

## Scope

Either:

- **Document (recommended, cheap):**
  - State chain-incapability in the module docs of `crates/bloqade-lanes-search/src/generators/greedy.rs`, naming the mechanism (`find_path_occupied` walls off all atoms) so the doc can't rot silently.
  - Pin it with a test on `crate::test_utils::chain_arch_json()` asserting the serialized behavior (follower moves, leader waits) — the pattern of the since-flipped `heuristic_selection_does_not_yet_assemble_chains`, so anyone who later makes greedy chain-capable is told to flip the assertion and update the docs rather than letting them drift.
- **Or implement** chain-following, replacing the single-atom BFS with joint pathfinding. Only worth considering if a real workload on a chain-capable spec shows greedy being selected and losing to the heuristic path.

## Non-goals

Changing greedy's spectator handling — never promoting a stationary atom to a mover is deliberate and shared with every other layer (#919).

PR #945, opened 2026-08-21
feat(benchmarks): add a conveyor-capable archspec suite

### LINKED ISSUE #939: Add a conveyor-capable spec to the benchmark registry to quantify chain assembly
Part of #887 (last scope bullet).

Chain assembly is dead code on both shipped Gemini specs — every bus is endpoint-disjoint, so the benchmark suites were byte-identical before/after #896 and #919, and CI cannot see a chain regression. The only quantification so far is ad hoc: #896 measured roughly a 20% move-count reduction on a 98-atom Ising kernel on an overlapping-bus spec, and #919 added `tests/conveyor_1d.rs` (correctness and optimality against a brute-force oracle) — but nothing in `python/benchmarks` exercises a chain-capable spec, so the win is neither tracked nor protected.

## Scope

- Add a synthetic conveyor-capable ArchSpec to the benchmark registry. The search crate's `chain_arch_json()` (example arch with site bus 0 rewired `0→1, 1→2, 2→3, 3→4`) is the ready-made shape; the benchmark likely wants a larger variant.
- Add at least one kernel/case where chains are profitable (a packed corridor / rigid-shift workload), across the strategies that now assemble chains (heuristic-generator-backed strategies and entropy).
- Regenerate and commit both baseline CSVs per the AGENT.md procedure; confirm determinism across two runs.
- Expected signal: fewer `move_count_events` / `move_count_lanes` (one AOD operation per chain instead of one per hop) at unchanged `success`.

## Notes

- `wall_time_ms` is excluded from the comparison; don't treat its noise as signal.
- Shipped-spec rows cannot form chains, so any diff outside the new rows is a red flag, not a baseline to accept.
- Pre-existing expected failures in the physical suite must stay put.

weinbe58:
Closes #939 (part of #887). Builds on #943, now merged.

Chain assembly was dead code on both shipped Gemini specs. A site bus is an
elementwise hop map `src[i] -> dst[i]`, and every shipped bus is
**endpoint-disjoint**, so `vacating_lane()` is necessarily `None` and every chain
path — #896's grid-layer repair and #919's `close_chain_entries` — is
unreachable. The benchmark suites were byte-identical before and after both PRs,
so CI could not see a chain regression.

## The spec

`examples/arch/gemini-conveyor.json` is the bundled Gemini physical spec with each
hypercube site bus converted to *conveyor* form. Dimension `d` has stride
`s = 2^d` and becomes `i -> i + s`:

| dim | stride | builtin lanes | conveyor lanes | overlap |
|---|---|---|---|---|
| 0 | 1 | `0→1, 2→3, 4→5, 6→7` | `0→1, 1→2, … 6→7` | `{1..6}` — one 8-long chain |
| 1 | 2 | `0→2, 1→3, 4→6, 5→7` | `0→2, 1→3, 2→4, 3→5, 4→6, 5→7` | `{2,3,4,5}` — two 4-long chains |
| 2 | 4 | `0→4, 1→5, 2→6, 3→7` | identical | `∅` — already disjoint, fixed point |

Each conveyor bus is a strict lane **superset** of the bus it replaces, and the
geometry is byte-identical, so a conveyor row differs from its builtin row only
by bus structure. Transport paths are purely **additive**: a `LaneAddr` is keyed
by source site id rather than an index into the bus `src` array, so all 1120
existing entries stay byte-identical and only 100 are added, synthesized in the
bundled spec's own waypoint idiom (`+5` y-clearance, uniform across all ten
site-bus words).

Generated, not hand-maintained — `just benchmark-conveyor` runs
`scripts/gen_conveyor_arch.py --check` first and fails if the committed JSON has
drifted from the bundled Gemini spec.

## Measured

81 rows against the physical baseline; 71 where both succeed:

| metric | better | same | worse |
|---|---|---|---|
| `move_count_events` | 71 | 0 | **0** |
| `move_count_lanes` | 67 | 4 | **0** |
| `estimated_fidelity` | 58 | 13 | **0** |
| `nodes_explored` | 59 | 3 | 9 |

Plus **4 new successes, 0 lost**. Mean −18.9% move events.

**Fidelity is the headline.** It is a product of per-gate fidelities, so it is
roughly exponential in operation count and a linear move reduction compounds:
`steane_physical_35` turned −19% move events into a **2.6× fidelity gain**.
Median +6% across rows with a nonzero baseline, and **9 of 22 underflow-floor
rows are lifted off `0.0`**.

The cost is confined to entropy — larger per-node candidate set, worst +139%
nodes — while distance-guided strategies get strictly *cheaper* (mean −23.7%),
because more lanes shorten graph distances and tighten the heuristic that guides
them. All nine more-expensive rows still produced better plans.

Note the superset property makes *plan existence* monotone, which is weaker than
"results can only improve" — every strategy is a bounded heuristic search. The
zero-regression outcome is empirical.

## What these numbers do NOT show, and the control that keeps it honest

The move-count gains do **not** isolate chain assembly. Chain-incapable
`rust_greedy` (#938) shows the same delta on every case measured, and on
`steane_physical_35` posts the best event count outright — so the win is the
richer lane set, not chains. Because chain-capability *is* the src/dst overlap,
no spec pair can hold the lane set fixed and toggle chains, so the two effects
cannot be separated by spec alone.

Therefore:

- **`rust_greedy` stays in the matrix as the connectivity-only control.** This is
  documented in AGENT.md with a "do not remove it" note.
- **Chain assembly is pinned by tests**, not by these metrics — see the
  `test(search)` commit here, and the existing coverage in `conveyor_1d.rs`.
- **No new kernel was added.** Its only purpose would have been a
  greedy-deadlock gate; skipping it keeps the shipped baselines byte-identical,
  since kernel discovery is global.

## Wiring

The CI benchmark job is already parameterized by `architecture`, so enabling this
is a one-word matrix change. One sharp edge, found by running it: `--output` is
**required** in the recipe. The default output path derives from `--architecture`
(`physical` here), not from `--compare`, so omitting it overwrites
`latest_physical.csv` with conveyor rows *while still reporting "no differences
found"* against the conveyor baseline.

## Verification

- Conveyor baseline stable across three consecutive runs, and re-verified after
  rebasing onto #937's motion-model change — no differences.
- `just benchmark-physical` and `just benchmark-logical`: no differences, so the
  shipped baselines are untouched.
- 463 Rust lib tests + `conveyor_1d` green; 81 Python arch/benchmark tests green;
  clean fmt/clippy/black/isort/ruff/pyright.
- Six guards in `python/tests/arch/test_conveyor_arch.py` pin the spec's defining
  properties: every site bus overlaps, each is a superset of its hypercube
  counterpart, geometry matches builtin exactly, existing paths are preserved
  byte-identically, and every site-bus lane has a transport path.

Rationale and the hardware questions this does *not* answer are in
`docs/superpowers/specs/2026-08-21-conveyor-benchmark-arch-design.md`.

## Follow-ups not in scope

- The entropy **deadlock-breaker** chain path (`drivers/entropy.rs:488`) still has
  no coverage, in particular the invariant that followers are exempt from the
  `target_movers` cap.
- #940 (`pack_aod_rectangles` missing the chain closure) is unaffected — the DSL
  strategy is deliberately excluded from the benchmark matrix, so this suite
  cannot see it.


weinbe58:
Closing this until move-generator rectangle size is bounded.

## Why

Neither the frontier (heuristic-generator) nor the entropy path puts an **upper
bound on the size of a generated AOD rectangle**. `EntropyParams` bounds the
*number* of candidates and movesets (`max_candidates`, `max_movesets_per_group`,
`max_goal_candidates`) but never the size of a single rectangle, and
`ops/aod_grid.rs` has no size limit anywhere in `greedy_init` / `try_add_point` /
`merge_clusters`.

That makes this suite's headline numbers unsafe to publish. A conveyor bus is a
strict lane superset, and chain assembly lets one rectangle span an entire
8-site chain — so the measured "fewer, larger operations" win can be bought with
rectangles the hardware may not be able to drive. Reporting a 2.6x fidelity gain
that partly comes from physically unrealizable shots would be worse than
reporting nothing.

## This is a pre-existing gap, not one this branch introduces

Worth stating plainly: the missing bound affects the **shipped** specs too. The
existing `latest_physical.csv` is produced by the same unbounded generators. What
this branch does is raise the ceiling and make the gap consequential — chain
assembly on overlapping buses is exactly the mechanism that would produce large
rectangles.

## What was measured before closing

Max lanes in a single `move.Move`, entropy vs astar:

```
builtin   ghz_6               max=2  mean=1.3
conveyor  ghz_6               max=2  mean=1.4
builtin   steane_physical_35  max=6  mean=1.6   dist: 2->16, 3->12, 4->6,  6->2
conveyor  steane_physical_35  max=6  mean=1.9   dist: 2->16, 3->2,  4->16, 6->2
```

So on the cases checked the conveyor spec does **not** currently blow up
rectangle size — the max is 6 on both — but it does shift mass toward larger
rectangles (4-lane shots go 6 -> 16). The risk is latent rather than currently
binding, which is an argument for fixing it deliberately rather than under
schedule pressure.

## Not lost

The branch `feat/conveyor-benchmark-arch` stays. The design doc
(`docs/superpowers/specs/2026-08-21-conveyor-benchmark-arch-design.md`) carries
the full rationale, the measured results, and the attribution analysis, and will
be updated with this blocker.

Reopen once rectangle size is bounded and the baseline is regenerated against
the bounded generators. #939 stays open.

weinbe58:
Addendum — the large cases finished measuring after I closed this, and they
change the framing.

Max lanes in a single `move.Move`, `rust_entropy_5`:

```
builtin   adder_64          n=1540  max=9  mean=1.2   4->62, 6->8,  9->2
conveyor  adder_64          n=1492  max=9  mean=1.2   4->50, 6->6,  9->6
builtin   trotter_rand_35   n=1184  max=9  mean=1.4   4->54, 6->4,  9->10
conveyor  trotter_rand_35   n=1050  max=9  mean=1.5   4->72, 6->18, 9->4
```

**The shipped spec already emits 9-lane rectangles today.** `adder_64` and
`trotter_rand_35` produce them on bundled Gemini, on `main`, with no conveyor
involved — 10 such shots in the `trotter_rand_35` builtin row. So the unbounded
rectangle problem is live and observable in the committed
`latest_physical.csv`, not something this branch would introduce.

The conveyor spec does not raise the ceiling (9 on both, 6 on both for
`steane_physical_35`); it shifts mass toward mid-size rectangles
(`trotter_rand_35` 6-lane shots 4 -> 18).

Practical consequence for the fix: whatever bound gets chosen has to be applied
to the **existing** baselines as well, and is likely to change them. That makes
bounding rectangle size a standalone piece of work with its own baseline
regeneration, rather than a prerequisite this branch merely trips over.

PR #946, opened 2026-08-21
fix: terminalmeasvalidation enforces exactly one terminal measure

jasonhan3:
- terminalmeasurevalidation enforces one terminal measure (errs if no terminal measure is found)


weinbe58 (approved):
lgtm just needs a test.

jasonhan3:
Should we be careful about backporting this because technically it might be "breaking" for users who define logical kernels without `terminal_measure`? Or is that expected

